PR #3, opened 2021-05-04
Backbone-module_CU-j3b2gk

a-wilkes:


a-wilkes:
Task linked: [CU-j3b2gk Backbone module](https://app.clickup.com/t/j3b2gk)

lpboucher:
Task linked: [CU-j3b2gk Backbone module](https://app.clickup.com/t/j3b2gk)

src/app/modules/backbone/backbone.common.ts
```diff
@@ -0,0 +1,27 @@
...
+import { AppComponent } from "@app/app.component";
+import { BackbonePageComponent } from "@modules/backbone/pages/backbone/backbone.page";
+import { BackboneComponent } from "@modules/backbone/components/backbone/backbone.component";
+
+export const pageDeclarations: any[] = [
+    BackbonePageComponent,
+];
+
+export const componentDeclarations: any[] = [
+    BackboneComponent
```
lpboucher: No need for a backbone component here, we will add other components such as header/sidebar, footer

src/app/modules/backbone/backbone.common.ts
```diff
@@ -0,0 +1,27 @@
...
+export const providerDeclarations: any[] = [
+];
+
+export const routes: Routes = [
+    {
+        path: "",
+        component: BackbonePageComponent,
+        children: [
+            { path: "", component: AppComponent },
+            { path: "backbone", component: BackboneComponent },
```
lpboucher: both of these routes are not needed: the appcomponent will not contain any html, but will rather load only a router-outlet. It is bootstrapped automatically on app load so there is no need to add it here.

The children route here will contain routes to components that will apread "around" the sidebar and the footer. They will components like product-list, product-detail etc.

lpboucher (changes requested):
you can remove the whole backbone component and leave children route array empty for now.

lpboucher (approved):
All good! merging

lpboucher:
merged

PR #6, opened 2021-05-05
Product-card-component_CU-j55m2b

a-wilkes:


a-wilkes:
Task linked: [CU-j55m2b Product card component](https://app.clickup.com/t/j55m2b)

lpboucher:
Task linked: [CU-j55m2b Product card component](https://app.clickup.com/t/j55m2b)

src/app/modules/products/product.interface.ts
```diff
@@ -0,0 +1,8 @@
+export interface Product {
+    id: number;
+    name: string;
+    origin: string;
+    price: string;
+    description: string;
+    tastingNotes: string;
+}
```
lpboucher: this interface is nice. I like that you moved it out of the component file.

It will eventually be turned into a class so that we can init instances of Product with data returned from an API. The class will then be used by services and components. Because it will be used it many places, we should locate it in the shared folder (under models directory)

src/app/modules/products/components/product-card/product-card.component.html
```diff
@@ -0,0 +1,40 @@
+<div class="product-card">
+    <div class="product-card-image-container">
+        <img class="product-card-image" src="https://images.unsplash.com/photo-1475855581690-80accde3ae2b?ixlib=rb-1.2.1&ixid=eyJhcHBfaWQiOjEyMDd9&auto=format&fit=crop&w=750&q=80" alt="">
+    </div>
+    <div class="product-card-body">
+        <span class="product-card-body-highlight">New</span>
+        <h2 class="product-card-body-title">{{product.name}}</h2>
+        <p class="product-card-body-description">{{product.description}}</p>
+        <div class="product-card-body-price-container">
```
lpboucher, line 9: this looks really nice. I removed the card footer because we will not be using the stars, but otherwise well done!

src/app/modules/products/components/product-card/product-card.component.scss
```diff
@@ -0,0 +1,71 @@
+.product-card {
```
lpboucher, line 1: Only change I made is to add a h-full to the product-card to force cards that are smaller to take up the full height of their container.

Otherwise, I split the classes into their categories (also added in the guidelines) for better readability

src/app/modules/products/components/product-card/product-card.component.ts
```diff
@@ -0,0 +1,21 @@
...
+import { Product } from '../../product.interface';
+
+@Component({
+  selector: 'app-product-card',
+  templateUrl: './product-card.component.html',
+  styleUrls: ['./product-card.component.scss']
+})
+export class ProductCardComponent implements OnInit {
+    @Input() product!: Product;
+
```
lpboucher, line 20: i have added two getters here (was not explained in the ticket so no worries). I am simply looking ahead when we add more product types and how their mapping might not be the same

lpboucher (approved):
Looks good!

We can start working with services at this point to really start building out some functionality

PR #9, opened 2021-05-10
Extend-product-service-with-featured-products_CU-jfb33c

a-wilkes:


a-wilkes:
Task linked: [CU-jfb33c Extend product service with featured products](https://app.clickup.com/t/jfb33c)

lpboucher:
Task linked: [CU-jfb33c Extend product service with featured products](https://app.clickup.com/t/jfb33c)

src/app/core/products/product.service.ts
```diff
@@ -0,0 +1,124 @@
...
+
+    getProducts(): Observable<Coffee[]> {
+        return of(this.products);
+    }
+
+    getFeaturedProducts(): Observable<Coffee[]> {
+        return of(this.products
+            .filter(p => p.collection === "featured")
+            .slice(0, 2));
+    }
```
lpboucher, line 111: this looks good in our case. However, we eventually want the data to be piped from the api directly in the filter. To make things easier for us, we should use the rxjs operator filter instead of the vanillajs array filter. That way, we can filter "on the fly" as values are received from an API. You can read up on it: https://www.learnrxjs.io/learn-rxjs/operators/filtering/filter

Here, we should call getProducts in getFeaturedProducts and then filter the observable (using rxjs) and return the resulting observable

src/app/shared/models/types/collection-type.type.ts
```diff
@@ -0,0 +1 @@
+export type CollectionType = "featured" | "upcoming" | undefined;
```
lpboucher, line 1: looks good

lpboucher (changes requested):
requesting changes to use rxjs operator instead of vanilla array method.

This is not something I would have expected you to figure out. I just realised when I was reviewing. In the end, we most likely will have an endpoint on the backend that will return these products, but in any case, it is good to get familiar with the rxjs operators (they are quite tricky)

src/app/core/products/product.service.ts
lpboucher: I also requested changes to ProductService, so you will have to rebase this branch after you made the changes

lpboucher (approved):

PR #10, opened 2021-05-10
Product refactor

a-wilkes:
Makes coffee and product types OO instead of procedural.

### DIFF
--- a/src/app/shared/models/classes/coffee.class.ts
+++ b/src/app/shared/models/classes/coffee.class.ts
@@ -1,0 +1,35 @@
+import { Product } from "./product.class"
+
+export class Coffee extends Product {
+    origin: string;
+    tastingNotes: string;
+    process: string;
+    varietal: string;
+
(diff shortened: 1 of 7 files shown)


src/app/shared/models/classes/coffee.class.ts
```diff
@@ -0,0 +1,35 @@
...
+    }
+
+    get displayedDescription() {
+        return [this.process, this.varietal];
+    }
+
+    get displayedDetails(): string[] {
+        return [this.origin, this.tastingNotes];
+    }
+}
```
lpboucher, line 38: I like it! Nice one

I will make a simple change. Basically, there is a high likelihood that the data received from the back-end will be incomplete (or some route will provide limited data to reduce data xfer in http requests) so we have to accoutn for incomplete objects passed to the constructor of the class. Typescript has a type called a Partial

lpboucher (approved):

PR #11, opened 2021-05-10
Extend-product-service-with-single-product-getter_CU-jfb8wq

a-wilkes:


a-wilkes:
Task linked: [CU-jfb8wq Extend product service with single product getter](https://app.clickup.com/t/jfb8wq)

src/app/shared/models/classes/product.interface.ts
```diff
@@ -6,4 +6,5 @@ export interface Product {
     price: string;
     description: string;
     type: ProductType;
+    slug: string;
```
lpboucher: nice, this should indeed be a property of product

src/app/core/products/product.service.ts
```diff
@@ -0,0 +1,122 @@
...
+    constructor() { }
+
+    ngOnInit() {
+    }
+
+    getProducts(): Observable<Coffee[]> {
+        return of(this.products);
+    }
+
+    getOne(slug: string): Observable<Coffee | undefined> {
```
lpboucher: Typescript complaining about result of find being undefined right?

Going forward, we will have an endpoint to getasingle product and we will route to a 404 page if the slug does not exist. So we can leave as is for now

lpboucher (changes requested):
Looks good, from my side.

Maybe you can rebase on the product-service after you have made the changes there

lpboucher (approved):

PR #12, opened 2021-05-11
Single-product-page_CU-jfb8tb

a-wilkes:


a-wilkes:
Task linked: [CU-jfb8tb Single product page](https://app.clickup.com/t/jfb8tb)

lpboucher:
Task linked: [CU-jfb8tb Single product page](https://app.clickup.com/t/jfb8tb)

src/app/modules/products/pages/single-product/single-product.page.html
```diff
@@ -0,0 +1,10 @@
+<div *ngIf="product | async as c">
```
lpboucher: app does not compile because product does not exist, should be product$

src/app/modules/products/pages/single-product/single-product.page.html
```diff
@@ -0,0 +1,10 @@
...
+    <h1>{{ c.name }}</h1>
```
lpboucher: We should use desriptive variable names.

Here we can either change c to currentProduct or change product$ to currentProduct$ and then c to product

src/app/modules/products/pages/single-product/single-product.page.ts
```diff
@@ -0,0 +1,32 @@
...
+
+import { Product } from "@shared/models/classes/product.class";
+
+@Component({
+    selector: "app-single-product",
+    templateUrl: "./single-product.page.html",
+    styleUrls: ["./single-product.page.scss"]
+})
+export class SingleProductPageComponent implements OnInit {
+    product$?: Observable<Product | undefined>;
```
lpboucher, line 14: I see why we need undefined as well. Realistically, when we request data from the server, we may actually redirect to a separate.page if the slug does not exist, so getOne should never return undefined.

Okay to leave as is for now, we will correct later

src/app/modules/products/pages/single-product/single-product.page.ts
```diff
@@ -0,0 +1,32 @@
...
+
+    constructor(
+        private route: ActivatedRoute,
+        private productService: ProductService,
+    ) { }
+
+    ngOnInit(): void {
+        this.route.paramMap
+            .subscribe(params => {
+                const slug = params.get("slug");
```
lpboucher: here we should use routeSnapshot to get params synchronously:

this.slug = this.route.snapshot.paramMap.get("slug");

see https://angular.io/api/router/ActivatedRouteSnapshot

lpboucher (changes requested):

lpboucher (approved):

PR #14, opened 2021-05-12
Featured-card-component_CU-jfb3rp

a-wilkes:


a-wilkes:
Task linked: [CU-jfb3rp Featured card component](https://app.clickup.com/t/jfb3rp)

lpboucher:
Task linked: [CU-jfb3rp Featured card component](https://app.clickup.com/t/jfb3rp)

src/app/core/products/product.service.ts
```diff
@@ -112,7 +112,9 @@ export class ProductService {
     getFeaturedProducts(): Observable<Product[]> {
         return this.getProducts()
             .pipe(
-                map(arr => arr.filter(p => p.collection === "featured")),
+                map(arr =>
+                    arr.filter(p => p.collection === "featured")
+                        .slice(0, 2)),
                 take(2)
```
lpboucher, line 118: do you need both slice(0,2) and take(2)?

src/app/modules/products/components/featured-product-card/featured-product-card.component.html
```diff
@@ -0,0 +1,16 @@
...
+        <div class="featured-product-card-body-header">
+            <h2 class="featured-product-card-body-title">{{product.name}}</h2>
+            <span class="featured-product-card-body-highlight">{{product.collection}}</span>
+        </div>
+        <p class="featured-product-card-body-description" *ngFor="let description of product.displayedDescription">{{description}}</p>
+        <div class="featured-product-card-body-price-container">
+            <div class="featured-product-card-body-price">{{product.price}}</div>
+            <button class="featured-product-card-body-get-notified" (click)="onClick()">Get notified</button>
+        </div>
+    </div>
```
lpboucher, line 15: I removed the bottom part just because we only want to show a teaser of the product and limit the space it takes

lpboucher (approved):
Nice one man!

Looks really good, and also responsive

src/app/core/products/product.service.ts
a-wilkes: So my thinking here was that we're returning an `Observable<Product[]>` which means we need to map the array down to two values using `slice`, but once the back-end is connected, we'll be returning an `Observable[Product]` at which point we need the `take` to return just two values. So yeah, the `take` is pointless at the moment because we only ever produce a single array, but later on, when the whole `map` bit gets removed, the `take` will come in to play and I figured it'd be prudent to leave it there for now so I don't forget about it further down the line. Maybe that's a bit cluttered, though?

PR #16, opened 2021-05-12
Login-guard_CU-hh6ywn

a-wilkes:


a-wilkes:
Task linked: [CU-hh6ywn Login guard](https://app.clickup.com/t/hh6ywn)

a-wilkes:
You mentioned both `canActivate` and `canLoad` in the todo list, but unless I misunderstood you only need the one implemented? I chose `canActivate`, but if that's wrong it's an easy fix. I also couldn't find any information in the documentation about `BaseGuard`, but it worked without it—let me know if it's something I need to bring in.

lpboucher:
Well, technically they do different things so it is likely that we will need both.

canActivate checks if the route can be activated, while canLoad checks if the module can be loaded. Realistically, as the app/module grows in size, we will not want to load it unnecessarily.

lpboucher:
No, you are right. There is specific requirement for a baseguard. The base guard is an abstract class that we can extend to add navigation capabilities to the guard. Basically, we will eventually have multiple guards that will all have some sort of related functionality that we can colocate in this abstract class to keep the guards lean

lpboucher:
I have made some minor changes, you can take a look.

Apologies, I realise now I was not all that clear.

I will request changes only to rebase once login form changes are done

lpboucher (changes requested):
Need to rebase

PR #17, opened 2021-05-17
Login-errors_CU-jx7kdu

a-wilkes:


a-wilkes:
Task linked: [CU-jx7kdu Login errors](https://app.clickup.com/t/jx7kdu)

lpboucher:
Task linked: [CU-jx7kdu Login errors](https://app.clickup.com/t/jx7kdu)

lpboucher (approved):
You were on the right track with the checks to show the errors.

You can review the changes I made to see how you can use the internal form state to display errors.

Basically, we use a general condition on the error container to see if we need to display any errors. And the individual error messages just need to check for the presence of errors in their own validation. It also allows to keep simple conditions in the view template, given that all conditions concern display. The component file remains lean and clean

PR #18, opened 2021-05-17
Extend-product-service-with-type-fetching_CU-jqanj7

a-wilkes:


a-wilkes:
Task linked: [CU-jqanj7 Extend product service with type fetching](https://app.clickup.com/t/jqanj7)

lpboucher:
Task linked: [CU-jqanj7 Extend product service with type fetching](https://app.clickup.com/t/jqanj7)

a-wilkes:
Since TS compiles down to JS and JS doesn't have types, I had to reintroduce `ProductType` which I don't feel is an ideal solution since it now means that there's a tight coupling between `ProductType` and the `Product` subclasses, as well as a redundancy in the information `ProductType` conveys. I'm not sure how else to do it, though, since you can't pass types as parameters. The only non-tightly coupled solution I could think of is a series of nearly identical getter methods like `getCoffees`, `getMerchandise` etc, which would only make sense if we can query the server-side API for these types (and even then, we'd couple the HTTP requests to the types). Am I over-thinking this, or am I missing a more elegant solution?

src/app/shared/models/classes/product.class.ts
```diff
@@ -7,6 +8,7 @@ export abstract class Product {
     collection: CollectionType | null = null;
     description: string | null = null;
     slug: string | null = null;
+    abstract productType: ProductType;
```
lpboucher: abstract properties should be at top of class

lpboucher: done

src/app/core/products/product.service.ts
```diff
@@ -118,4 +119,15 @@ export class ProductService {
                 take(2)
             );
     }
+
+    getProductsByType(productType?: ProductType | "all"): Observable<Product[]> {
+        if (productType === undefined || productType === "all") {
```
lpboucher: undefined gets marshalled to false, so we can use the bang (!) operator

lpboucher: done

lpboucher (approved):

lpboucher:
Sorry, just saw this.

No, you have a good point! You are getting into more advanced Typescript features.

There are a few things we can do. Let’s discuss at a certain point.

You can check the data-api.service.ts and see how the return types include a dynamic parameter type (Observable<T>).

PR #19, opened 2021-05-18
Merchandisecoffees-route_CU-jq54gw

a-wilkes:


a-wilkes:
Task linked: [CU-jq54gw Merchandise/coffees route](https://app.clickup.com/t/jq54gw)

lpboucher:
Task linked: [CU-jq54gw Merchandise/coffees route](https://app.clickup.com/t/jq54gw)

src/app/modules/products/components/product-list/product-list.component.html
```diff
@@ -1,14 +1,5 @@
...
-            <div class="product-card-featured-container" *ngFor="let product of featuredProducts$ | async">
-                <app-featured-product-card [product]="product"></app-featured-product-card>
-            </div>
-        </div>
-        <div class="product-list-items-container">
-            <div class="product-card-container" *ngFor="let product of products$ | async">
-                <app-product-card [product]="product"></app-product-card>
-            </div>
-        </div>
+<div class="product-list-items-container">
```
lpboucher: i think in this case it would be preferable, to leave the "product-list-items-container" in product-page.

This element does not need to know how its container is positioning it.

src/app/modules/products/pages/product/product.page.html
```diff
@@ -1 +1,8 @@
-<app-product-list></app-product-list>
+<div class="product-page-container">
+    <div class="product-page-featured-items-container">
+        <div class="product-page-featured-items-card-container" *ngFor="let product of featuredProducts$ | async">
+            <app-featured-product-card [product]="product"></app-featured-product-card>
+        </div>
+    </div>
+    <app-product-list [products$]=products$></app-product-list>
```
lpboucher: there is no need to pass the full observable here, we can extract the array before passing it to the list component.

Observables should be defined in the same component where the data is fetched. Otherwise, if we use @input, we can safely assume that the data is present by the time we reach that point in the template. You can also use the async pipe when passing data:

[products]="products$ | async"

src/app/modules/products/components/product-list/product-list.component.ts
```diff
@@ -11,13 +11,10 @@ import { Product } from "@shared/models/classes/product.class";
     styleUrls: ["./product-list.component.scss"]
 })
 export class ProductListComponent implements OnInit {
-    featuredProducts$: Observable<Product[]> = new Observable();
-    products$: Observable<Product[]> = new Observable();
+    @Input() products$: Observable<Product[]> = new Observable();
```
lpboucher: no need for an observable, this should be:

@Input() products: Product[] = [];

src/app/modules/products/pages/product/product.page.ts
```diff
@@ -1,15 +1,29 @@
 import { Component, OnInit } from "@angular/core";
+import { Observable } from "rxjs";
+
+import { ProductService } from "@app/core/products/product.service";
+
+import { Product } from "@app/shared/models/classes/product.class";
```
lpboucher: can use route aliases to shorten (@shared, @core)

src/app/core/products/product.service.ts
```diff
@@ -4,6 +4,8 @@ import { take, map } from "rxjs/operators";
 
 import { Coffee } from "@shared/models/classes/coffee.class";
 import { Product } from "@shared/models/classes/product.class";
+import { ProductType } from "@app/shared/models/types/product-type.type";
+import { Merchandise } from "@app/shared/models/classes/merchandise.class";
```
lpboucher: route aliases (@shared)

lpboucher (changes requested):

src/app/modules/products/components/product-list/product-list.component.html
lpboucher: All good

src/app/modules/products/pages/product/product.page.html
lpboucher: all good

lpboucher (approved):

PR #20, opened 2021-05-18
Sidebar-routing_CU-jqaqd8

a-wilkes:


a-wilkes:
Task linked: [CU-jqaqd8 Sidebar routing](https://app.clickup.com/t/jqaqd8)

lpboucher:
Task linked: [CU-jqaqd8 Sidebar routing](https://app.clickup.com/t/jqaqd8)

a-wilkes:
Ok, so this one ended up taking a while (mostly due to css stuff not playing along) and turned out to maybe be more of a change that you were perhaps intending. Essentially, to process the new navigation list under *constants*, I though it would be better to make a new component so that it could handle all the logic w.r.t. things being clicked/focussed etc. instead of trying to put this into the HTML for the sidebar. Hopefully it looks ok, though I'm not sure how I feel about the submenus—it might be better to have the parents of submenus only open and close the submenu, instead of having them link to a parent route (I did it this way since I thought you'd want a page that links to all products, but you didn't mention it in the items you wanted as children of the main products page in the todo list. You might think of a better way of doing this, if so let me know—I'm happy to keep working on it!)

src/app/modules/backbone/components/nav-item/nav-item.component.scss
```diff
@@ -0,0 +1,47 @@
...
+    @apply rotate-180;
+}
+
+.nav-item-arrow-closed {
+    @apply nav-item-arrow;
+    @apply rotate-0;
+}
+
+.nav-item-children-container {
+    @apply w-1/2 z-50;
```
lpboucher: can remove the w-1/2

src/app/modules/backbone/components/nav-item/nav-item.component.html
```diff
@@ -0,0 +1,14 @@
...
+    <span *ngIf="children.length > 0">
+        <svg [ngClass]="{'nav-item-arrow-open': isOpen, 'nav-item-arrow-closed': !isOpen}" fill="currentColor" viewBox="0 0 20 20">
+            <path fill-rule="evenodd" d="M5.293 7.293a1 1 0 011.414 0L10 10.586l3.293-3.293a1 1 0 111.414 1.414l-4 4a1 1 0 01-1.414 0l-4-4a1 1 0 010-1.414z" clip-rule="evenodd">
+            </path>
+        </svg>
+    </span>
+</li>
+<div [ngClass]="{'nav-item-children-container-open': isOpen, 'nav-item-children-container-closed': !isOpen}">
+    <ul *ngFor="let child of children">
+        <app-nav-item [rawNavData]="child" [parentLink]="link" (click)="onClickChild()"></app-nav-item>
```
lpboucher: instead of passing the whole object, we should pass individual properties:

@Input label: ...
@Input link: ...

That way, we can pass link to children as `${parent}${child}

Also makes it more readable

src/app/modules/backbone/components/nav-item/nav-item.component.scss
```diff
@@ -0,0 +1,47 @@
+.nav-item {
+    @apply p-2 rounded-lg;
+    @apply no-underline;
+
+    @apply hover:text-gray-800 hover:bg-gray-100 transition-colors;
+}
+
+.nav-item-active {
+    @apply underline;
```
lpboucher: can apply the same style as when hovering (see line5)

src/app/modules/backbone/components/nav-item/nav-item.component.ts
```diff
@@ -0,0 +1,38 @@
...
+
+import { NavigationItemType } from "@utils/constants/navigation";
+
+@Component({
+    selector: "app-nav-item",
+    templateUrl: "./nav-item.component.html",
+    styleUrls: ["./nav-item.component.scss"]
+})
+export class NavItemComponent implements OnInit {
+    @Input() rawNavData!: NavigationItemType;
```
lpboucher: pass individual properties of the object

src/app/modules/backbone/components/nav-item/nav-item.component.ts
```diff
@@ -0,0 +1,38 @@
...
+    @Input() parentLink: string = "";
+
+    label: string = "";
+    link: string = "";
+    children: NavigationItemType[] = [];
+    isOpen = false;
```
lpboucher: isOpen should likely be defined in the sidebar component

lpboucher (changes requested):
Well done man! This was really good. The comments I have are mainly because it was done well!

I think generally, we do not really need the click handlers in the navitem component. We can achieve the same behaviors by opening if link is active or on hover.

To do that, we can use pseudo-selectors:
1) target sibling of .nav-item-active + .nav-item-children-container-open
2) .nav-item:hover + .nav-item-children-container-open

Happy to discuss/explain. Well done!

Probably need to rebase on develop

lpboucher:
I was expecting it would take a bit more time. It is quite common, but is actually quite complex in css.

I agree with the use of another component. I think my description of the ticket was in no man’s land… a little too detailed to do only the css, and not detailed enough to really get into the full functionality.

You did really well though. I have added a lot of comments so we can review together, will be a good learning experience!

a-wilkes:
Ok, so I should have made all the changes we spoke about, the only issue is that I couldn't work out how to close the menus in the mobile version of the site when the user is scrolling. Any ideas? (also note that the little menu arrows are still there, I've just hidden them when the menu is forced open/currently active since it's a bit confusing as you expect to be able to click it to close the menu, and obviously you can't).

lpboucher:
Nice one. I am accepting this.

There is a bit of a bug at the moment (will create a ticket to fix). If you click on one of the sub-menus, you see the route will blink but revert back to /products.

Let's fix it in another ticket

PR #22, opened 2021-05-25
Filter-component_CU-jx8mqm

a-wilkes:


a-wilkes:
Task linked: [CU-jx8mqm Filter component](https://app.clickup.com/t/jx8mqm)

lpboucher:
Task linked: [CU-jx8mqm Filter component](https://app.clickup.com/t/jx8mqm)

a-wilkes:
I think this is everything we talked about yesterday. Remind me if I forgot something.

- I didn't bring in the multiselect-dropdown yet, but can do if you want.
- I had to hard-code the filter pipe so it only works for coffees at the moment. I'm not sure how you want to fix this up since it's not very extensible, but I'd love to hear some ideas about it.
- The filter component needs to be given a `FilterType` as input so it knows what to show to the user, we'll need to add some logic for this somewhere, but I don't know what the most sensible place would be.

Let me know what needs fixed with this one, as I'm not sure it's good to go in its current state.

src/app/shared/pipes/filter.pipe.ts
```diff
@@ -0,0 +1,38 @@
...
+
+@Pipe({
+    name: "filter"
+})
+export class FilterPipe implements PipeTransform {
+
+    transform(products: Product[], filter: FilterType): Product[] {
+        let filteredProducts = [];
+
+        const coffees = products
```
lpboucher: I know you had a comment about this, but the pipe should not be aware of the type of data structure it is receiving.

it should receive an arbitrary array of objects (object[]), along with an activeFilter (it is correct the way you have done it).

Then, we should (using the array.filter method):
1) loop through the keys of the object, check if the current key is included in the keys of the array of activeFilters
2) if it is found, it should check whether the value at that key in included in the activeFilters array at that same key (the value should check for a string or for an array of strings)

see below an example for a pipe that filters based on a searchString. The concept is similar, we just have to extend it a bit (note the first two lines)
`
import { Pipe, PipeTransform } from "@angular/core";

@Pipe({
    name: "filter"
})
export class FilterPipe implements PipeTransform {

    transform(items: object[], searchText: string): object[] {
        if (items == null) { return []; }
        if (searchText == null) { return items; }

        searchText = searchText.toLowerCase();

        return items.filter(item =>
            Object.keys(item).some((key) => {
                return typeof item[key] === "string" && item[key].toLowerCase().includes(searchText);
            })
        );
    }
}
`

lpboucher (changes requested):

lpboucher:
This was a good try. It is a bit ore difficult than I remembered XD

Here are some clarifications:
1) the pipe should not know what kind of data is passed or what fields are filterable. It should only process the action of filtering
2) We can implement the https://github.com/NileshPatel17/ng-multiselect-dropdown. It will be much easier to handle the styling and callbacks

Let's leave it on hold for now, we can discuss it together

a-wilkes:
Ok, this branch should now be rebased onto the latest develop, with a much more sensible filter pipe, and using ng-multiselect-dropdown.
- I didn't add styles yet since it's already styled by default and I figure we'll probably change it when we know the final design of the site.
- The filter and pipe are not currently used anywhere since you never mentioned it in the ticket and I understood it to be the case that we'll maybe be adding a service to read the URL queries and filtering that way. If not, let me know and I can add it in.

lpboucher (approved):

lpboucher:
Nice one! I will approve since it all works well.

I still think we are overcomplicating things, but it will make more sense once we have a filter service or more functionality, so lets leave it for now and we can revisit later

PR #24, opened 2021-05-25
Products-module-routing_CU-k99mvh

a-wilkes:


a-wilkes:
Task linked: [CU-k99mvh Products module routing](https://app.clickup.com/t/k99mvh)

a-wilkes:
So it turns out the problem wasn't with clashing names (though I've still fixed it up as described), but due to the children menu items getting moved inside the div of the parent—which was controlling the router link. What ended up happening was the router would get directed to the child's URL, then also trigger for the parent URL right after.

Unfortunately, fixing everything up as described meant the component wasn't getting reloaded when you visited different routes, so I had to subscribe (and unsubscribe!) to the URL parameters changing inside the product page component. I couldn't see a way to do this with aync pipes in the HTML, but as always I'm open to fixing with better solutions!

lpboucher (approved):
Works well.

Your comment was correct. We have to subscribe to params to listen to route changes because the component is technically the same, its output is different.

I have made some minor changes:
- if we subscribe within a component, it is good practice to use a single "subscriptions" variable that is private and that holds all observables
- we then use subscriptions.add to subcribe to different observables
- we can also use a single unsubscribe

I have also simplified the param getter: using params is simpler than paramMap as it allows us to deconstruct our parameter in the function callback.

PR #25, opened 2021-05-26
Gridlist-component_CU-j56bn1

a-wilkes:


a-wilkes:
Task linked: [CU-j56bn1 Grid/list component](https://app.clickup.com/t/j56bn1)

lpboucher:
Task linked: [CU-j56bn1 Grid/list component](https://app.clickup.com/t/j56bn1)

src/app/modules/products/components/product-list/product-list.component.html
```diff
@@ -1,3 +1,5 @@
-<div class="product-card-container" *ngFor="let product of products">
-    <app-product-card [product]="product"></app-product-card>
-</div>
+<app-grid-list [mobileColumnCount]=1>
+    <div class="product-card-container" *ngFor="let product of products">
```
lpboucher: this div is not needed. We can use the ngFor loop and pass the product on the app-product-card element

By removing this intermediary div, we can more easily target children of the grid item. Also, the product-card-container class is not needed (see comment below)

src/app/modules/products/components/product-list/product-list.component.scss
```diff
@@ -1,5 +1,3 @@
 .product-card-container {
-    @apply w-full;
     @apply p-4;
```
lpboucher: this class is not needed. Here, adding padding around the individual elements in the grid should be the responsibility of the grid layout. It may seem counterintuitive because it has to be applied on each element, but you can use grid-gap for gap between elements in a grid

src/app/modules/products/pages/product/product.page.html
```diff
@@ -1,10 +1,10 @@
...
+        <div class="featured-item-container" *ngFor="let product of featuredProducts$ | async">
             <app-featured-product-card [product]="product"></app-featured-product-card>
         </div>
-    </div>
+    </app-grid-list>
     <div class="product-list-container" *ngIf="products$ | async as products">
-        <app-product-list class="product-list" [products]="products"></app-product-list>
+        <app-product-list [products]="products"></app-product-list>
     </div>
 </div>
```
lpboucher, line 10: Here it is becoming a bit confusing (I just noticed). featured-product-card is wraped by the grid-list while product-list uses the grid-list internally.

I think we can now remove the product-list and simply use the grid-list here (as we do for featured-product-card)... anyway , i will create another ticket

src/app/modules/products/pages/product/product.page.scss
```diff
@@ -1,23 +1,12 @@
...
 
-.product-page-featured-items-container {
-    @apply flex flex-wrap;
-    @apply -m-4;
-}
-
-.product-page-featured-items-card-container {
-    @apply w-full;
+.featured-item-container {
     @apply p-4;
```
lpboucher, line 7: these intermediary classes/div are not needed. But we can fix in another ticket

src/app/shared/components/grid-list/grid-list.component.ts
```diff
@@ -0,0 +1,36 @@
...
+    @Input() itemColumnSpan: number = 1;
+
+    constructor() { }
+
+    ngOnInit(): void {
+        this.mobileColumnCount = this.clampColumnValue(this.mobileColumnCount);
+        this.itemColumnSpan = this.clampColumnValue(this.itemColumnSpan);
+    }
+
+    getColumnClass(): string {
```
lpboucher: if a method takes in no parameters and only derives value from another property, we should use class getters.

src/app/shared/components/grid-list/grid-list.component.ts
```diff
@@ -0,0 +1,36 @@
...
+        return `mobile-columns-${ this.mobileColumnCount }`;
+    }
+
+    getItemColumnSpan(): string {
+        return `item-span-${ this.itemColumnSpan }`;
+    }
+
+    private clampColumnValue(columnValue: number): number {
```
lpboucher, line 40: I like this method. It seems likely that we will use it some other place correct?

This is the kind of transformation that is a great use case for a pipe, pipes can also be used in component files by injecting them in the constructor and calling the transform method. I will explain in another ticket

lpboucher (changes requested):
Some minor changes needed. This was well done

src/app/shared/components/grid-list/grid-list.component.ts
```diff
@@ -24,6 +25,18 @@ export class GridListComponent implements OnInit {
...
+        const baseString = "grid-gap-";
+        const validGapValues = [0, 0.5, 1, 1.5, 2, 2.5, 3, 3.5, 4, 5, 6, 7, 8, 9, 10,
+            11, 12, 14, 16, 20, 24, 28, 32, 36, 40, 44, 48, 52, 56, 60, 64, 72, 80, 96];
+
+        if (validGapValues.includes(this.gridGap)) {
+            return baseString + this.gridGap.toString().replace(".", "-");
+        }
+
+        return baseString + "0";
+    }
```
lpboucher, line 38: this is a bit of an overkill. We are only applying classes here, so any invalid values will not throw errors or bugs it simply will not apply any styles

simple string interpolation will do

lpboucher: done

lpboucher (approved):

PR #26, opened 2021-05-31
Wallet-widget_CU-knbfrj

a-wilkes:


a-wilkes:
Task linked: [CU-knbfrj Wallet widget](https://app.clickup.com/t/knbfrj)

a-wilkes:
I'm using the wallet icon from Google fonts, this is under the Apache 2.0 licence. If we want to keep using it, we'll need to add the license to the repo, but I'm not sure what kind of business impact that has for you, so I can remove it if needs be.

lpboucher (approved):

PR #30, opened 2021-06-01
Floating-samples-button_CU-kqcqvy

a-wilkes:


a-wilkes:
Task linked: [CU-kqcqvy Floating samples button](https://app.clickup.com/t/kqcqvy)

src/app/modules/backbone/components/floating-action/floating-action.component.html
```diff
@@ -0,0 +1,3 @@
+<div class="floating-action-background">
```
lpboucher: why the container div? Can be removed

You only a button that uses fixed position

src/app/modules/backbone/components/floating-action/floating-action.component.scss
```diff
@@ -0,0 +1,12 @@
+.floating-action-background {
+    @apply fixed flex items-center justify-center z-50;
+    @apply w-full h-full;
+    @apply bg-gray-600 bg-opacity-70;
+}
+
```
lpboucher: background class is not needed, move the fixed class to the button

src/app/modules/backbone/components/floating-action/floating-action.component.scss
```diff
@@ -0,0 +1,12 @@
...
+.floating-action-button {
+    @apply p-4 rounded-full shadow cursor-pointer;
```
lpboucher: should add fixed class with right-10, bottom-10 to position in lower right corner of the screen

please rearrange classes with new lines for layout, sizing, presentations, pseudo (hover)

lpboucher (changes requested):
Minor changes.

Please make sure you always include the component somewhere on the page even if just for testing purposes

src/app/modules/backbone/components/floating-action/floating-action.component.html
a-wilkes: Ah ok, I think I misunderstood—I added in the background so that I could dim the rest of the page when the button appears. I will remove it.

src/app/modules/backbone/components/floating-action/floating-action.component.scss
a-wilkes: Yes, this clarifies it for me—I thought it was supposed to be like an alert in the centre of the screen that dims everything else, but in fact it should be a little pop-up in the corner!

lpboucher: Ahhh makes sense why you had it that way initially.

Yes, exactly. The floating button should be visible at (almost) all times.

lpboucher (approved):

PR #33, opened 2021-06-07
Cart-service_CU-knbjcw

a-wilkes:


a-wilkes:
Task linked: [CU-knbjcw Cart service](https://app.clickup.com/t/knbjcw)

lpboucher:
Task linked: [CU-knbjcw Cart service](https://app.clickup.com/t/knbjcw)

a-wilkes:
~~The good news is that because we're using `snipcart.store.getState().cart.items`, we don't need to wait for `document.Snipcart.ready` anymore, so that solves that problem (at least for now).~~

Thought this was the case, turns out it's not, however I think I've found a solution which you can see in the [Cart-widget PR](https://github.com/lpboucher/dak-coffee-wholesale/pull/34#discussion_r647241122). Let me know how you want to integrate it (or not, as the case may be).

src/app/core/cart/cart.service.ts
```diff
@@ -0,0 +1,14 @@
...
+
+@Injectable({
+    providedIn: 'root'
+})
+export class CartService {
+    private snipcart: any = (window as any).Snipcart;
+
+    constructor() { }
+
+    get cartTotal(): number {
```
lpboucher, line 9: getters should be placed above the constructor, below the class properties

lpboucher: done

lpboucher (approved):

PR #34, opened 2021-06-08
Cart-widget_CU-knbhrw

a-wilkes:


a-wilkes:
Task linked: [CU-knbhrw Cart widget](https://app.clickup.com/t/knbhrw)

lpboucher:
Task linked: [CU-knbhrw Cart widget](https://app.clickup.com/t/knbhrw)

src/app/core/cart/cart.service.ts
```diff
@@ -0,0 +1,47 @@
...
+    }
+
+    ngOnDestroy() {
+        this.subscriptions.unsubscribe();
+    }
+
+    private evaluateCartWeight(): number {
+        const cartItems = this.snipcart.store.getState().cart.items.items;
+        return cartItems
+            .filter((item: any) => item?.dimensions?.weight != null && item?.quantity != null)
```
a-wilkes, line 44: At the moment, this filters everything as none of the items in the cart ever have a non-null weight attribute. I assume this is something that will be fixed when we add a drop-down for picking the weight in the future?

src/app/core/cart/cart.service.ts
```diff
@@ -0,0 +1,47 @@
...
+    private static readonly REFRESH_RATE: number = 1000;
+
+    private subscriptions: Subscription = new Subscription();
+    private snipcart: any = {};
+    cartTotal$: Observable<number> = new Observable();
+    cartWeight$: Observable<number> = new Observable();
+
+    constructor() {
+        this.subscriptions.add(
+            fromEvent(document, "snipcart.ready")
```
a-wilkes, line 17: This looks like the way to subscribe to this event. It definitely works, at least! Do you still want to make a service for this as you mentioned with `EventManager`, or is it better to just do this where we need to?

src/app/shared/components/cart-widget/cart-widget.component.scss
```diff
@@ -0,0 +1,19 @@
...
+    @apply text-indigo-500;
+
+    @apply md:mb-1 md:mx-auto md:block;
+}
+
+.cart-contents {
+    @apply text-gray-900 inline;
+
+    @apply md:font-medium md:text-xl;
+}
```
a-wilkes, line 19: This is now identical to the wallet widget (along with most of the HTML). Should we abstract this down into shared widget component?

src/app/core/cart/cart.service.ts
lpboucher: I have implemented a solution, I will explain

src/app/core/cart/cart.service.ts
lpboucher: Yes, that's fine for now

src/app/shared/components/cart-widget/cart-widget.component.scss
lpboucher: that's a good idea. let's see if we figure out a solution

lpboucher (approved):

PR #35, opened 2021-06-08
Dynamic-modal-content-projection_CU-kzb3m2

a-wilkes:


a-wilkes:
Task linked: [CU-kzb3m2 Dynamic modal content projection](https://app.clickup.com/t/kzb3m2)

lpboucher:
Task linked: [CU-kzb3m2 Dynamic modal content projection](https://app.clickup.com/t/kzb3m2)

lpboucher:
Hey man, looks good!

I just glanced quickly and I am missing the ng-content call inside the modal-component. Before I review, we need to implement it in a fully dynamic way.

The content of the modal, should be in a new component: samples-confirmation-modal. As in the example, it should be wrapped by app-modal.

Essentially, form now on, we want to be able to declare new modals by simply defining their content and making them children of app-modal

src/app/modules/backbone/components/samples-confirmation-modal/samples-confirmation-modal.component.html
```diff
@@ -0,0 +1,17 @@
+<app-modal [display]="display">
```
a-wilkes: This is a bit weird since it looks like we're binding `ModalComponent`'s `display` attribute to itself. Which isn't exactly right. What's going on is that the modal fades out when it's closed, but if the css class which controls the opacity is looking at the `display` attribute of its respective component (i.e. without this binding), this will be the `display` attribute of the component that Angular creates when it sees `<app-modal>` in the HTML of `SamplesConfirmationModalComponent` (`SCMComponent`), which is a different `display` attribute from the one in `SCMComponent` even though `SCMComponent` is a subclass of `ModalComponent`. Which I admit is confusing. I've tried to explain a little better in the diagram below.

Without binding `display`:
```
<div [class.open]="display" ...> ... <!-- In shared/.../modal.component.html-->
                       \
ModalComponent          \  ModalComponent (created by Angular when it sees <app-modal> in modules/.../samples-confirmation-modal.component.html)
┌─────────┐              \  ┌─────────┐
│ display │               ->│ display │
└─────────┘                 └─────────┘
     ^
     | extends
     |
SCMComponent (created by us when the user clicks on the floating action button)
┌─────────┐
│...      │
└─────────┘
````

However, when we do this binding, we ask `ModalComponent` to look at the `display` attribute of the `SCMComponent` we created, not its own one (the one created by Angular).

File: src/app/core/modals/modal.service.ts [hunk 2]
```diff
@@ -0,0 +1,37 @@
...
+        if (this.currentModelRef) { return; }
+
+        this.currentModelRef = this.componentFactoryResolver
+            .resolveComponentFactory(component)
+            .create(this.injector);
+
+        this.applicationRef.attachView(this.currentModelRef.hostView);
+
+        const domElem = (this.currentModelRef.hostView as EmbeddedViewRef<any>)
+            .rootNodes[0] as HTMLElement;
```
lpboucher: can use array desctructuring to remove reference to [0]

lpboucher: done

lpboucher (approved):
Really nice man, this works really well.

I think the result is pretty bang on: proper separation of concerns, good reusability, pretty DRY code

PR #39, opened 2021-06-09
Pricing-service_CU-m18zaq

a-wilkes:


a-wilkes:
Task linked: [CU-m18zaq Pricing service](https://app.clickup.com/t/m18zaq)

src/app/core/pricing/pricing-tier.service.ts
```diff
@@ -0,0 +1,16 @@
+import { Injectable } from "@angular/core";
+import { BehaviorSubject, Observable } from "rxjs";
+
+@Injectable({
+    providedIn: "root"
+})
+export class PricingTierService {
+    private isVolumeDiscountActive$ = new BehaviorSubject(false);
+    toggleDiscount = () => this.isVolumeDiscountActive$.next(!this.isVolumeDiscountActive$.value);
```
a-wilkes, line 9: Using a getter resulted in `undefined` errors cropping up over the place due to the context of `this` ([more info here](https://stackoverflow.com/a/57588867)).

a-wilkes: Use ```Output``` to pass value from toggle upwards.

lpboucher: I made the fix using @Output.

You can check out the code on the develop branch if you want to see how its done. I also removed the observable from the toggle-widget component. That way, the component remains "dumb" and only received true or false and emits a new value

PR #41, opened 2021-06-14
Use-actual-api_CU-m18rxy

a-wilkes:


a-wilkes:
Task linked: [CU-m18rxy Use actual api](https://app.clickup.com/t/m18rxy)

lpboucher:
Task linked: [CU-m18rxy Use actual api](https://app.clickup.com/t/m18rxy)

src/app/core/products/coffee-api.service.ts
```diff
@@ -22,10 +22,7 @@ export class CoffeeApiService extends DataApiService<Coffee> {
     }
 
     getCoffee(slug: string): Observable<Coffee | undefined> {
```
lpboucher: i think we can remove the undefined type

src/app/core/products/merchandise-api.service.ts
```diff
@@ -1,36 +1,25 @@
...
+    constructor(protected http: HttpClient) {
+        super(config.backendURL + "wholesale/", http, Merchandise);
+    }
 
     getMerchandise(): Observable<Merchandise[]> {
-        return of(this.merchandise);
+        return this.getAll("merchandises?isActive=true");
     }
 
     getMerchandiseItem(slug: string): Observable<Merchandise | undefined> {
```
lpboucher: i think we can remove the undefined type

src/app/core/products/product-api.service.ts
```diff
@@ -26,12 +26,8 @@ export class ProductApiService extends DataApiService<Product> {
...
 
     getProducts(): Observable<Product[]> {
-        return forkJoin([
-            this.coffeeApiService.getCoffees(),
-            this.merchandiseApiService.getMerchandise(),
-        ]).pipe(
-            map((products: Product[][]) => flatten(products)),
-        );
+        return zip(this.coffeeApiService.getCoffees(), this.merchandiseApiService.getMerchandise())
+            .pipe(map((products: Product[][]) => flatten(products)));
```
lpboucher, line 30: i agree with you this is hard to understand. Let's revisit at a later point. I think we can one/a combination of rxjs operators that do a better job of conveying what is happening here

src/app/core/products/product-api.service.ts
```diff
@@ -51,13 +47,20 @@ export class ProductApiService extends DataApiService<Product> {
     }
 
     getProductsByType(productType?: ProductType | "all"): Observable<Product[]> {
-        if (!productType || productType === "all") {
-            return this.getProducts();
+        switch (productType) {
+            case undefined: /* fallthrough */
```
lpboucher: is this needed?

fallthrough will just fall to the default case no? Also, the default case in this example should be all

In any case, use a hashmap. Switch statements have O(n) complexity, while objects (with unique keys) are O(1)

I prefer) use an object (e.g dict) where the keys are the product types (all, coffee, etc.) the values are the returned observables.

And then simply return dict[productType] || dict["all"].

lpboucher (changes requested):

lpboucher (approved):

PR #42, opened 2021-06-14
Product-click_CU-md8k1m

a-wilkes:


a-wilkes:
Task linked: [CU-md8k1m Product click](https://app.clickup.com/t/md8k1m)

src/app/modules/products/components/product-card/product-card.component.html
```diff
@@ -1,4 +1,4 @@
-<div class="product-card">
+<div class="product-card" (click)="onClick($event)">
```
lpboucher, line 1: no need to use a click event, you can simply use routerLink here

src/app/modules/products/components/product-card/product-card.component.html
```diff
@@ -8,7 +8,7 @@ <h2 class="product-card-body-title">{{product.name}}</h2>
         <p *ngFor="let description of product.displayedDescription">{{description}}</p>
         <div class="product-card-body-price-container">
             <div class="product-card-body-price">{{product.price}}</div>
-            <button
+            <button #addToCartButton
```
lpboucher, line 11: this is not needed

src/app/modules/products/components/product-card/product-card.component.ts
```diff
@@ -8,10 +9,19 @@ import { Product } from "@shared/models/classes/product.class";
...
     ngOnInit(): void {
     }
+
+    onClick(event: Event): void {
+        const elementUnderClick = event.target as HTMLElement;
+        const addToCartButtonElement = this.addToCartButton.nativeElement as HTMLElement;
+        if (elementUnderClick == addToCartButtonElement) { return; }
+
+        this.router.navigate(["products", this.product.productType, this.product.slug]);
+    }
```
lpboucher, line 26: this (although a nice solution) is not needed. A simple routerLink in the html will do

lpboucher (changes requested):

src/app/modules/products/components/product-card/product-card.component.ts
a-wilkes: As discussed this morning, I (just re-)tried the router link with `stopPropagation`, `stopImmediatePropagation`, and `preventDefault` on the click event for the "Add to cart" button, and they either don't stop the router link from activating, or prevent snipcart from seeing the click. It looks like the alternative is `Snipcart.api.cart.items.add({...})` which feels like it undermines the directive a little since we wouldn't need to bind the product details to the element's attributes anymore as we could just pass the details straight into the snipcart API from the component. Should I leave this for now and we can make a new ticket later to create an "addToCart" method in the snipcart service, or do you have an idea of something else I could try?

a-wilkes:
Closing in favour of new PR/ticket: [Product-click_CU-ptc01j](https://github.com/lpboucher/dak-coffee-wholesale/pull/95).

PR #43, opened 2021-06-14
Extend-cart-service_CU-knbmzh

a-wilkes:


a-wilkes:
Task linked: [CU-knbmzh Extend cart service](https://app.clickup.com/t/knbmzh)

a-wilkes:
As mentioned, there are still some issues with functionality:
- The `snipcart.initialized` event does not always fire/is not always caught
- If `snipcart.initialized` does fire, the cart widget does not necessarily read the next values from the `BehaviourSubject`s (although the correct values are pushed with `next`).
- The cart widget does not see the new values sent by `next` when adding, updating, or removing items from the cart

src/app/app.component.ts
```diff
@@ -12,25 +12,27 @@ import { SnipcartEvents } from "@shared/models/types/snipcart-events.type";
 export class AppComponent implements OnInit, OnDestroy {
     private subscriptions: Subscription = new Subscription();
     title = "dak-wholesale";
-    cartEvents!: SnipcartEvents;
+    cartEvents: SnipcartEvents = defaultSnipcartEvents();
```
lpboucher: I am not a big fan of this. Colocating types and functions... let's leave as it was previously. Unless you have a specific reason?

lpboucher (changes requested):
To fix the race condition, there is also an event that we can subscribe to for any changes to the snipcart state: Snipcart.store.subscribe.

Maybe we can try it to see if is more stable

src/app/app.component.ts
a-wilkes: Originally, in *@shared/models/types/snipcart-events.type.ts* you had
```
export type SnipcartEvents = {
    addingItemSubscription: () => {},
    addedItemSubscription: () => {},
    updatedItemSubscription: () => {},
    removedItemSubscription: () => {},
    orderCompletedSubscription: () => {}
};
```
And you were using it as seen on the diff for this comment chain, which was causing `ERROR TypeError: Cannot set property 'addedItemSubscription' of undefined` since `cartEvents` never had an object assigned to it. Assigning an empty object wouldn't work (`cartEvent: SnipcartEvents = {}`) since the empty object `{}` is missing the defined properties (`addingItemSubscription` etc.), so somewhere in the code there needed to be a default object with all the properties defined to empty functions, hence the change in *@shared/models/types/snipcart-events.type.ts* to:
```
export type SnipcartEvents = {
    snipcartInitializedSubscription: () => void,
    addingItemSubscription: () => void,
    addedItemSubscription: () => void,
    updatedItemSubscription: () => void,
    removedItemSubscription: () => void,
    orderCompletedSubscription: () => void,
};

export function defaultSnipcartEvents(): SnipcartEvents {
    return {
        snipcartInitializedSubscription: () => {},
        addingItemSubscription: () => {},
        addedItemSubscription: () => {},
        updatedItemSubscription: () => {},
        removedItemSubscription: () => {},
        orderCompletedSubscription: () => {},
    };
}
```
I did it this way since I thought it was cleaner than populating *app.component.ts* with the default definition. Is there an alternative that fixes the problem? Happy to change to whatever is best.

a-wilkes:
I tried out the `Snipcart.store.subscribe` method and it seems to work fine, but if we use this it would mean getting rid of/changing most of `snipcart.service.ts` since we would no longer be waiting for specific cart events, right? Arguably, this might make more sense at the moment since we do exactly the same thing no matter what the user's action is, however, I assume this will change in the future, so maybe it wouldn't make sense going forward. Do you want me to leave as is for now (with possible change requests coming from the `SnipcartEvents` discussion above)?

src/app/app.component.ts
lpboucher: Ahhh I see where the issue is... That was a mistake on my part, when the types were defined as () => {}, it was meant as a function that returns nothing (the curly braces were supposed to identify the function block not the return type).

Complete brain fuck on my part. You are right that we can need to update SnipcartEvents so all functions are void, but there is no need for the defaultSnipcartEvents method since we never need to return empty objects

lpboucher:
Let's add the store.subscribe in snipcart service and call a method to update the cart in cart.service. That method should just call .next on the different behaviour subjects that we define (number of items, weight) etc.

In this case, we will be calling next even if the value has not changed, which is okay because change detection does not run if the value does not change

You are right that the store is all we need to listen to for deriving updates to our UI. However, we can also keep the individual cart event blocks (leave them empty or logging) as these will eventually be used to trigger Google anaytics reporting.

lpboucher:
So to recap, we should:

1) add the store subscribe method to snipcart service, which calls updateCart, passing in the relevant pieces of state we will be using (items)
2) in cart service, for every update call .next with the new current value for the number of items and the cart total
3) keep the cart events in snipcart service but simply log their result
4) if snipcart.initialized does not add any benefit to fixing the race condition (over store subscribe), remove the event listener

I made the changes to the snipcart-events type file so make sure you pull before adding your changes

a-wilkes:
Obviously, the cart widget still doesn't update correctly all the time, but this is the same problem as was fixed in [Toggle-logic](https://github.com/lpboucher/dak-coffee-wholesale/pull/47#discussion_r657150199). If the Toggle-logic solution is ok with you, then I can bring it in for the cart widget as well (I didn't do it here since I figure it's a different issue, so it might be cleaner to keep it apart).

a-wilkes:
Obviously, the cart widget still doesn't update correctly all the time, but this is the same problem as was fixed in [Toggle-logic](https://github.com/lpboucher/dak-coffee-wholesale/pull/47#discussion_r657150199). If the Toggle-logic solution is ok with you, then I can bring it in for the cart widget as well (I didn't do it here since I figure it's a different issue, so it might be cleaner to keep it apart).

PR #47, opened 2021-06-14
Toggle-logic_CU-md8j1q

a-wilkes:


a-wilkes:
Task linked: [CU-md8j1q Toggle logic](https://app.clickup.com/t/md8j1q)

lpboucher:
Task linked: [CU-md8j1q Toggle logic](https://app.clickup.com/t/md8j1q)

src/app/core/pricing/pricing-tier.service.ts
```diff
@@ -5,6 +5,8 @@ import { BehaviorSubject, Observable } from "rxjs";
     providedIn: "root"
 })
 export class PricingTierService {
+    private static ITEM_COUNT_THRESHOLD: number = 4;
```
lpboucher: instead of use a static property, just use a const variable defined outside the class

src/app/core/pricing/pricing-tier.service.ts
```diff
@@ -16,4 +18,10 @@ export class PricingTierService {
     toggleDiscount(value?: boolean): void {
         this.isVolumeDiscountActive$.next(value ?? !this.isVolumeDiscountActive$.value);
     }
+
+    updateDiscount(itemCount: number): void {
```
lpboucher, line 26: where is this method called? the toggle method is not updated because the discount is not updated.

updateDiscount should be called from the cartService on updates

lpboucher (changes requested):

a-wilkes:
I rebased onto [Extend-cart-service_CU-knbmzh](https://github.com/lpboucher/dak-coffee-wholesale/pull/43) to call the `updateDiscount` method in `@core/cart/cart.service.ts` since it didn't really make sense without it (as previously on this branch we weren't hooked in to the cart events). I hope that's not too messy.

lpboucher (changes requested):
the toggle does not work as expected.

If I start with an empty cart and add 5 items. I should see the toggle change to 45% automatically when adding the fifth item.

It works on page load but not while the app is running.

Most likely an issue with change detection

lpboucher:
I just realised this should be done using reactive forms as well:

Check this example:
https://www.concretepage.com/angular-2/angular-2-radio-button-and-checkbox-example#reactive-form

We should define a formControl for the toggle, set the value using either setValue or patchValue.

If you need to check when the value of one of the inputs changes (isChecked), there is a lifecycle method called ngOnChanges that gets called everytime an input/output/observable changes

src/app/modules/backbone/components/sidebar/sidebar.component.ts
```diff
@@ -1,28 +1,39 @@
...
         private pricingTierService: PricingTierService,
+        private changeDetectorRef: ChangeDetectorRef,
     ) {}
 
     ngOnInit(): void {
-        this.priceTierActive$ = this.pricingTierService.isDiscountActive$;
+        this.subscriptions.add(
+            this.pricingTierService.isDiscountActive$.subscribe(value => {
+                this.priceTierActive = value;
+                this.changeDetectorRef.detectChanges();
```
a-wilkes, line 43: Of all the things I tried, this was the only thing that worked. Everything else failed to have the sidebar (and toggle) detect that any changes had occurred.

lpboucher: yes, I think you are right. I think this is a bit of an edge case where we need change detection to trigger from a subscription in the parent to the child...

As far as I understand is that the subscriptions marks the child as "needing a change" but does not run change detection.

All good, this is an acceptable solution otherwise we would have to create a standalone component that itself listens to updates to isDiscountActive$, which I do not want to do at this point

lpboucher: Well done!

PR #53, opened 2021-06-16
Related-products-list_CU-mfczva

a-wilkes:


a-wilkes:
Task linked: [CU-mfczva Related products list](https://app.clickup.com/t/mfczva)

src/app/shared/components/grid-list/grid-list.component.html
```diff
@@ -1,3 +1,3 @@
-<div class="grid-list-base" [ngClass]="[columnClass, itemColumnSpanClass, gridGapClass]">
+<div class="grid-list-base" [ngClass]="[columnClass, mdColumnClass, itemColumnSpanClass, gridGapClass]">
```
lpboucher: all the changes to the grid-list component are liekly not needed because the behaviour should be the same as the regular grid-list: as the screen size gets smaller, the number of columns should also adapt

As it is now, if you shrink the size of the screen, the columns are unreadable because their width is too small

See how the regular grid-list behaves and apply the same behaviour here: as the screen gets smaller, go down to 2 products per row, and then eventually one product per row (arranged in a column)

lpboucher (changes requested):

PR #54, opened 2021-06-16
Skeleton-loaders_CU-mhdp9g

a-wilkes:


a-wilkes:
Task linked: [CU-mhdp9g Skeleton loaders](https://app.clickup.com/t/mhdp9g)

lpboucher:
Task linked: [CU-mhdp9g Skeleton loaders](https://app.clickup.com/t/mhdp9g)

a-wilkes:
I had to put the HTML and CSS in the pages and not the respective components since the pages deal with the async pipes and the components just receive the values. Switching this around (at least for the product and featured product cards) seems awkward at best, since the observables unwrap to an array of products. We'd need to restructure all the product services if we wanted an array of observables of products to allow the components to handle the async pipes instead. What do you think?

src/app/modules/products/pages/product/product.page.html
```diff
@@ -1,8 +1,36 @@
 <div class="product-page-container">
     <app-grid-list [itemColumnSpan]=2 [gridGap]=4>
-        <app-featured-product-card *ngFor="let product of featuredProducts$ | async" [product]="product"></app-featured-product-card>
+        <ng-container *ngIf="featuredProducts$ | async as featuredProducts; else featuredProductsLoading">
+            <app-featured-product-card *ngFor="let product of featuredProducts" [product]="product"></app-featured-product-card>
+        </ng-container>
```
lpboucher, line 10: this is a good start.

the ngx-skeleton-loader and its ngIf should be in the component that it replaces. Here, in featured-product-card, we should check whether the product is available and if not, display the loader. Same for product-card and product-detail

The page components should not know whether the result will be a loader or an actual component, they should only be concerned about what will take up that specific space in the layout

src/app/modules/products/pages/product/product.page.html
```diff
@@ -1,8 +1,36 @@
...
+                    'width': '100%',
+                    'height': '150px',
+                    'border-radius': '0.5rem'
+                }">
+            </ngx-skeleton-loader>
+        </ng-template>
     </app-grid-list>
     <app-grid-list [gridGap]=4>
-        <app-product-card *ngFor="let product of products$ | async" [product]="product"></app-product-card>
+        <ng-container *ngIf="products$ | async as products; else productsLoading">
```
lpboucher: as above, move to product-card

src/app/modules/products/pages/single-product/single-product.page.html
```diff
@@ -1,3 +1,36 @@
-<div *ngIf="product$ | async as p">
-    <app-product-detail [product]="p"></app-product-detail>
+<div *ngIf="product$ | async as product; else productLoading">
+    <app-product-detail [product]="product"></app-product-detail>
```
lpboucher: as above, move to product-detail

src/app/modules/products/products.module.ts
```diff
@@ -1,20 +1,23 @@
-import { NgModule } from '@angular/core';
-import { CommonModule } from '@angular/common';
+import { NgModule } from "@angular/core";
+import { CommonModule } from "@angular/common";
 import { SharedModule } from "@shared/shared.module";
 
-import { ProductsRoutingModule } from './products-routing.module';
+import { NgxSkeletonLoaderModule } from "ngx-skeleton-loader";
```
lpboucher: these loaders may be used for other modules as well. Therefore, they should be imported in the shared module

lpboucher (changes requested):
This is a very good start! We jsut need to tweak it a bit:

See here:
https://user-images.githubusercontent.com/1252570/50053778-d4e0d900-018e-11e9-9de7-fad6f9fddd9e.gif

notice how the facebook skeleton exmaple combines multiple loaders (circle, full width etc.) to give the appearance that the loader knows what data it is expecting.

We should so the same. E.g. for product-card, our loader should have a large square at the top (as a placeholder for the image), then a few short ones (as placeholders for titles, arietal etc), two short ones (for price and add to card button), etc.

src/app/modules/products/components/featured-product-card/featured-product-card.component.html
```diff
@@ -1,16 +1,80 @@
-<div class="featured-product-card">
-    <div class="featured-product-card-image-container">
-        <img class="featured-product-card-image" src="https://images.unsplash.com/photo-1475855581690-80accde3ae2b?ixlib=rb-1.2.1&ixid=eyJhcHBfaWQiOjEyMDd9&auto=format&fit=crop&w=750&q=80" alt="">
+<ng-container *ngIf="product != null; else showLoader">
```
lpboucher: I dont like how there are two templates and two ngIfs (one in product-page and one in the card components

Feels like we are checking the same condition twice.

Let's move all the html for the loaders in their own components and then add them in the product-page:

`
<app-grid-list [itemColumnSpan]=2 [gridGap]=4>
        <ng-container *ngIf="featuredProducts$ | async as featuredProducts; else featuredProductsLoading">
            <app-featured-product-card *ngFor="let product of featuredProducts" [product]="product"></app-featured-product-card>
        </ng-container>
        <ng-template #featuredProductsLoading>
            <app-featured-product-card-loader></app-featured-product-card-loader>
        </ng-template>
    </app-grid-list>
`

lpboucher (changes requested):

lpboucher (approved):
Nicely done. I realized that we may run into some issues if our observable return falsy values (edge case) so I added a loader pipe that we can use to check if the observable is loading or has an error.

Skeletons look really good

PR #56, opened 2021-06-21
Register-method_CU-mtarx3

a-wilkes:


a-wilkes:
Task linked: [CU-mtarx3 Register method](https://app.clickup.com/t/mtarx3)

src/app/shared/models/classes/new-customer.class.ts
```diff
@@ -0,0 +1,32 @@
+type Sector = "cafe" | "restaurant" | "office" | "subscription box" | "reseller";
```
lpboucher: we should define this in a separate type file

src/app/shared/models/classes/new-customer.class.ts
```diff
@@ -0,0 +1,32 @@
...
+
+export class NewCustomer {
+    contactName: string = "";
+    businessName: string = "";
+    email: string = "";
+    password: string = "";
+    sector: Sector | undefined;
+    vatNumber: string | undefined;
+
+    constructor(
```
lpboucher: this class should use the same structure as the other classes we have defined:

1) it should use a newCustomerShape in the constructor that is of type Partial<NewCustomer>
2) all class variables should be initialized as null, false, [], {}
3) setting the initial values is done inside an if block: 
`if (newCustomerShape != null) {
            if (newCustomerShape.id != null) {
                this.id = newCustomerShape.id;
            }
}
`

lpboucher (changes requested):

lpboucher (approved):

PR #59, opened 2021-06-22
Modals-refactoring_CU-mt95zj

a-wilkes:


a-wilkes:
Task linked: [CU-mt95zj Modals refactoring](https://app.clickup.com/t/mt95zj)

lpboucher:
Task linked: [CU-mt95zj Modals refactoring](https://app.clickup.com/t/mt95zj)

src/app/modules/backbone/backbone.common.ts
```diff
@@ -7,7 +7,7 @@ import { SidebarComponent } from "./components/sidebar/sidebar.component";
 import { FooterComponent } from "./components/footer/footer.component";
 import { NavItemComponent } from "./components/nav-item/nav-item.component";
 import { FloatingActionComponent } from "./components/floating-action/floating-action.component";
-import { SamplesConfirmationModalComponent } from "./components/samples-confirmation-modal/samples-confirmation-modal.component";
+import { SamplesConfirmationModalComponent } from "@shared/components/modals";
```
lpboucher: this component should be exported from the shared module, like other modals

When we export components from a module, they become available in other modules. That is why we do not need to import them in multiple places and it is also why we use this shared module

lpboucher (changes requested):

lpboucher (approved):

PR #60, opened 2021-06-22
Request-access-form_CU-mtarfc

a-wilkes:


a-wilkes:
Task linked: [CU-mtarfc Request access form](https://app.clickup.com/t/mtarfc)

src/app/modules/authentication/components/request-access-form/request-access-form.component.ts
```diff
@@ -0,0 +1,111 @@
...
+            && errors.hasOwnProperty("notMatching")
+            && errors.notMatching === true
+            && this.shouldShowErrors(this.passwordConfirmControl);
+    }
+
+    private shouldShowErrors(control: AbstractControl): boolean {
+        return control.dirty || control.touched || this.submissionAttempted;
+    }
+
+    private control(name: string): AbstractControl | null {
```
lpboucher: i find this a bit redundant. It is sufficient to call requestAccessForm.get in the getters...

src/app/modules/authentication/components/request-access-form/request-access-form.component.ts
```diff
@@ -0,0 +1,111 @@
...
+        return this.requestAccessForm.get(name);
+    }
+
+    private matchPasswordsValidator(): ValidatorFn {
```
lpboucher, line 96: this is really good.

What we can do in Angular is define our own validators and pass them in the form control init array

I will create another ticket and we can refactor.

src/app/modules/authentication/components/request-access-form/request-access-form.component.scss
```diff
@@ -0,0 +1,62 @@
+.request-access-background {
+    @apply flex flex-col;
+    @apply h-screen;
```
lpboucher: using h-screen limits the background height to the height of the screen.

Here, since the form is longer than the height of the screen, we have a white bar at the bottom.

Instead, the request-access container should derive its height from the height of its child.

On mobile, there is a margin at the top. We should keep it for all devices and have the same at the bottom.

lpboucher (changes requested):
I have updated import statements, make sure you pull before making changes

lpboucher (approved):

PR #65, opened 2021-06-28
Custom-password-validator_CU-n5781p

a-wilkes:


a-wilkes:
Task linked: [CU-n5781p Custom password validator](https://app.clickup.com/t/n5781p)

src/app/modules/authentication/components/request-access-form/request-access-form.component.ts
```diff
@@ -0,0 +1,97 @@
...
+            sector: this.sectorControl.value,
+            vatNumber: this.vatNumberControl.value,
+        });
+    }
+
+    hasErrors(control: AbstractControl): boolean {
+        return control.invalid && this.shouldShowErrors(control);
+    }
+
+    hasUnmatchedPasswords(): boolean {
```
lpboucher: when a simple function takes in no parameters and returns a single (relatively simple value), we should use getters

Methods called in html templates will force change detection meaning in this case, the component will rerender every single which can get pretty costly performance-wise.

Also, we can simplify the boolean chain by using: errors?.notMatching && this.shouldShowErrors.

Basically, if errors.notMatching === true, it means that the errors != null and that notMatching exists so they are redundant

lpboucher: done

lpboucher (approved):

PR #66, opened 2021-06-29
Refactor-sidebar-and-header_CU-mxe2gm

a-wilkes:


a-wilkes:
Task linked: [CU-mxe2gm Refactor sidebar and header](https://app.clickup.com/t/mxe2gm)

src/app/modules/backbone/components/header/header.component.html
```diff
@@ -0,0 +1,20 @@
...
+        <app-burger-menu [onBurgerMenuClick]="onBurgerMenuClick"></app-burger-menu>
+    </div>
+    <div class="nav-spacer"></div>
+    <div class="widget-container">
+        <app-wallet-widget></app-wallet-widget>
+        <app-cart-widget></app-cart-widget>
+    </div>
+    <div class="sidebar-container" [ngClass]="{ 'sidebar-open': showSidebar, 'sidebar-closed': !showSidebar }">
+        <div class="sidebar">
+            <app-sidebar [(open)]="showSidebar"></app-sidebar>
```
lpboucher: by importing app-sidebar here, we are creating a second instance of the sidebar component. There is no need to do this. Also, there should be no dependency between the header/sidebar but rather they should communicate through input/output (if we need a service to facilitate, we can also do that)

src/app/modules/backbone/components/header/header.component.html
```diff
@@ -0,0 +1,20 @@
+<nav class="main-nav">
+    <header class="brand">
+        <img src="/favicon.ico"/>
+        <span>DAK</span>
+    </header>
+    <div class="burger-menu-wrapper">
+        <app-burger-menu [onBurgerMenuClick]="onBurgerMenuClick"></app-burger-menu>
```
lpboucher: we should always prefer to define @Output in the child component and handle the event here. It is bad practice to pass function references

src/app/modules/backbone/components/header/header.component.html
```diff
@@ -0,0 +1,20 @@
...
+    </div>
+    <div class="nav-spacer"></div>
+    <div class="widget-container">
+        <app-wallet-widget></app-wallet-widget>
+        <app-cart-widget></app-cart-widget>
```
lpboucher: we should keep using the generic widget component

src/app/modules/backbone/components/header/header.component.ts
```diff
@@ -0,0 +1,28 @@
...
+import { PricingTierService } from "@core/pricing/pricing-tier.service";
+
+import { SidebarComponent } from "@modules/backbone/components/sidebar/sidebar.component";
+
+@Component({
+    selector: "app-header",
+    templateUrl: "./header.component.html",
+    styleUrls: ["./header.component.scss"]
+})
+export class HeaderComponent extends SidebarComponent {
```
lpboucher: why extend sidebar? did you want to share some of the functionality?

It is not necessarily a bad idea and I am willing to try it out. However, if we want to go this route, for consistency, we should define a base abstract menu component with the shared functionality and have both sidebar and header extend from that base class

src/app/modules/backbone/components/nav-item/nav-item.component.html
```diff
@@ -1,5 +1,5 @@
-<li class="nav-item-container" routerLinkActive="nav-item-active" [ngClass]="{'nav-item-force-open': label == openLabel}">
-    <div class="nav-item" [routerLink]="[link]">
+<li class="nav-item-container" [ngClass]="{'nav-item-force-open': label == openLabel}">
+    <div class="nav-item" [routerLink]="[link]" routerLinkActive="nav-item-active">
         <span>{{ label }}</span>
         <span *ngIf="children.length > 0">
             <svg class="nav-item-arrow" fill="currentColor" viewBox="0 0 20 20">
```
lpboucher, line 5: We have lost the nesting of the menu items? did it happen in another merge? or is it a result of the change here?

src/app/modules/backbone/components/sidebar/sidebar.component.html
```diff
@@ -14,22 +18,11 @@
...
-                <path d="M0 0h24v24H0z" fill="none"/>
-                <path d="M21 18v1c0 1.1-.9 2-2 2H5c-1.11 0-2-.9-2-2V5c0-1.1.89-2 2-2h14c1.1 0 2 .9 2 2v1h-9c-1.11 0-2 .9-2 2v8c0 1.1.89 2 2 2h9zm-9-2h10V8H12v8zm4-2.5c-.83 0-1.5-.67-1.5-1.5s.67-1.5 1.5-1.5 1.5.67 1.5 1.5-.67 1.5-1.5 1.5z"/>
-            </svg>
-        </app-widget>
-        <app-widget class="widget-item" content="€{{ cartTotal | async }} ({{ cartWeight | async }} kg)" (widgetClickEvent)="onCartWidgetClicked()">
-            <svg xmlns="http://www.w3.org/2000/svg" height="24px" viewBox="0 0 24 24" width="24px" fill="currentColor">
-                <path d="M0 0h24v24H0V0z" fill="none"/>
-                <path d="M15.55 13c.75 0 1.41-.41 1.75-1.03l3.58-6.49c.37-.66-.11-1.48-.87-1.48H5.21l-.94-2H1v2h2l3.6 7.59-1.35 2.44C4.52 15.37 5.48 17 7 17h12v-2H7l1.1-2h7.45zM6.16 6h12.15l-2.76 5H8.53L6.16 6zM7 18c-1.1 0-1.99.9-1.99 2S5.9 22 7 22s2-.9 2-2-.9-2-2-2zm10 0c-1.1 0-1.99.9-1.99 2s.89 2 1.99 2 2-.9 2-2-.9-2-2-2z"/>
-            </svg>
-        </app-widget>
```
lpboucher: we should retain the existing generic widget

src/app/modules/backbone/components/wallet-widget/wallet-widget.component.ts
```diff
@@ -0,0 +1,18 @@
+import { Component } from "@angular/core";
```
lpboucher: this component is not needed, can be deleted

src/app/modules/backbone/components/sidebar/sidebar.component.html
```diff
@@ -3,6 +3,10 @@
         <img class="" src="/favicon.ico"/>
         <span class="">DAK</span>
     </header>
+    <div class="mobile-close" (click)="onMobileCloseClicked()">
+        <div class="mobile-close-bar-1"></div>
+        <div class="mobile-close-bar-2"></div>
+    </div>
```
lpboucher, line 9: why not use the burger menu component here?

src/app/shared/components/burger-menu/burger-menu.component.ts
```diff
@@ -0,0 +1,10 @@
+import { Component, Input } from "@angular/core";
+
+@Component({
+    selector: "app-burger-menu",
+    templateUrl: "./burger-menu.component.html",
+    styleUrls: ["./burger-menu.component.scss"]
+})
+export class BurgerMenuComponent {
+    @Input() onBurgerMenuClick: () => void = () => {};
```
lpboucher: use @Output instead. That way, instead of passing down the click handler, we can handle the callback in the parent component

lpboucher (changes requested):
This is a good start. Some comments:

- There should be no dependency between header and sidebar. Neither should handle a piece of state of the other (show[...] for example
- There should only one instance of sidebar
- Header and sidebar should not be responsible for their own visibility (apart from css classes). That should be the responsibility of the parent component as it is the parent that is responsible for layout
- All the logic that is now in header/sidebar can be moved into the parent backbone and passed down as inputs. Then, the sidebar and header can use @Output
- If you prefer inheritance over composition, then we should create an abstract menu class that will be extended by both header and sidebar

We will take the time to discuss this next time as there are lots of comments and it is a great example of best practices, single-responsibility, etc.

I have also merged develop into this branch so that all recent changes are added

src/app/modules/backbone/components/nav-item/nav-item.component.html
a-wilkes: This was deliberate. When we were applying the active class to the parent list-item, it highlighted the whole container which extended behind the children and caused an ugly appearance when the children were selected (highlighted in red to make it easier to see):
image

By highlighting the inner div, which doesn't include the children or the container which extends behind them, I think the result looks nicer (or, at least, less like a mistake. You may want to change it further, e.g. margins):
image

a-wilkes: I just realised you might be seeing the effect of the parent menu being forced open on activation. Without this, we'd be highlighting a menu item (e.g. products/coffees) that wasn't visible to the user unless they moused over the menu (potentially leading them to think they were looking at all products instead of just coffees, for example).

The little menu arrow also gets hidden when its respective menu gets forced open since I thought it was a little confusing to have an arrow indicating the presence of a mouse-over menu, but the menu was already forced open (meaning the arrow wasn't adding anything, just a redundant—potentially even confusing—detail).

src/app/modules/backbone/components/sidebar/sidebar.component.html
a-wilkes: You mean give the burger menu an animation to transition into a close-cross so it can fill both roles?

Wouldn't it be confusing for the user to click the sidebar closed, only for the close button to turn into a burger menu during the drawer closing animation? But maybe I'm envisaging it wrong—could you explain in more detail?

src/app/modules/backbone/components/nav-item/nav-item.component.html
lpboucher: Yes, I see what you mean. The dropdown works correctly if I navigate to general-information. So all good

src/app/modules/backbone/components/sidebar/sidebar.component.html
lpboucher: Yes you are right. I confused the close icon with the burger icon

lpboucher (approved):
Well done! The result is really clean with responsibility isolated

PR #71, opened 2021-07-05
Product-detail-refactor_CU-n7dy9j

a-wilkes:


a-wilkes:
Task linked: [CU-n7dy9j Product detail refactor](https://app.clickup.com/t/n7dy9j)

src/app/modules/products/components/product-detail/product-detail.component.ts
```diff
@@ -1,23 +1,60 @@
...
     templateUrl: "./product-detail.component.html",
     styleUrls: ["./product-detail.component.scss"]
 })
-export class ProductDetailComponent implements OnInit {
+export class ProductDetailComponent {
     @Input() product!: Product;
 
+    readonly weightOptions: Weight[] = ["250g", "1kg"];
+    readonly defaultWeight: Weight = this.weightOptions[0];
+    currentWeight: Weight = this.defaultWeight;
```
lpboucher: this code is duplicated across components. The arrays of options should be moved to a constant file.

FYI, here is a cool trick to generate string union type from an array of strings:
`
const weightOptions = ["250g", "1kg"] as const;

export type Weight = typeof weightOptions[number];
`

so we should define the array of options in a const file (form-options.ts), then import the weight options and generate the type from the array. Solves duplicating the options in an array and a type, but also gives us type completion (should also be done in product-card

src/app/modules/products/components/product-detail/product-detail.component.ts
```diff
@@ -1,23 +1,60 @@
...
+
+    readonly roastOptions: Roast[] = ["Filter", "Espresso", "Both"];
+    readonly defaultRoast: Roast = this.roastOptions[0];
+    currentRoast: Roast = this.defaultRoast;
+
+    snipcartOptions: CustomOption[] = this.makeSnipcartOptions();
+
+    selectionForm = this.fb.group({
```
lpboucher, line 29: the form is not used. We have two options:

1) we use form Controls for the radio buttons, then subscribe to onValueChanges and updage the value of snipcartOptions on form changes
2) we remove the form from the component and template and use the methods here. If you go this route, I would combine the onSelectWeight and onSelectRoast methods into one. You can pass a generic option variable. The currentWeight is only used in the options array. So it can all be done at once

lpboucher (changes requested):

src/app/modules/products/components/product-detail/product-detail.component.ts
lpboucher: The same should be done in the product-card

a-wilkes: This all changed in the [Add-quantity-input-to-product PR](https://github.com/lpboucher/dak-coffee-wholesale/pull/74) where I think it got refactored to something much cleaner than what I have here. Should I just leave that to be merged in later, or would you like me to bring it in here as well?

lpboucher (changes requested):
I think it is getting a bit confusing here. So I will request changes, but mainly to integrate changes

PR #72, opened 2021-07-06
Quantity-increment-custom-component_CU-nhbdr1

a-wilkes:


a-wilkes:
Task linked: [CU-nhbdr1 Quantity increment custom component](https://app.clickup.com/t/nhbdr1)

lpboucher:
Task linked: [CU-nhbdr1 Quantity increment custom component](https://app.clickup.com/t/nhbdr1)

src/app/shared/components/quantity-input/quantity-input.component.ts
```diff
@@ -0,0 +1,114 @@
...
+            multi: true,
+            useExisting: QuantityInputComponent,
+        }
+    ]
+})
+export class QuantityInputComponent implements ControlValueAccessor, Validator {
+    @Input() label: string = "";
+    @Input() maxValue?: number;
+    @Input() minValue?: number;
+    readonly defaultValue = 0;
```
lpboucher: no need for this readonly defaultvalue, 0 is pretty straightforward

src/app/shared/components/quantity-input/quantity-input.component.ts
```diff
@@ -0,0 +1,114 @@
...
+    _quantity: number = this.defaultValue;
```
lpboucher: can remove the _, and use 1 as initial value

src/app/shared/components/quantity-input/quantity-input.component.ts
```diff
@@ -0,0 +1,114 @@
...
+        {
+            provide: NG_VALIDATORS,
+            multi: true,
+            useExisting: QuantityInputComponent,
+        }
+    ]
+})
+export class QuantityInputComponent implements ControlValueAccessor, Validator {
+    @Input() label: string = "";
+    @Input() maxValue?: number;
```
lpboucher: I like the optional input, but it would make more sense to define default values that define the base behaviour we want. So here, maybe max is 20 by default, and minValue should be 0

src/app/shared/components/quantity-input/quantity-input.component.ts
```diff
@@ -0,0 +1,114 @@
...
+        this.touched = true;
+    }
+
+    private sanitizeValue(value: number): number {
+        return isNaN(value)
+            ? this.defaultValue
+            : this.clampToRange(value);
+    }
+
+    private clampToRange(value: number): number {
```
lpboucher: I dont believe these two methods are required:

1) sanitizing is not required because we define a custom number value and all types only take numbers as inputs so it is redundant to check for isNan
2) clampToRange already exists as a pipe. We should instead import the pipe in this component and call .transform in the setter

lpboucher (changes requested):

src/app/shared/components/quantity-input/quantity-input.component.ts
a-wilkes: The reason I check for NaN is because the user can type their own value into the input element in the HTML. If they type characters, nothing will be entered into the box, but it will return a blank value to the code here and cause the browser to spit out an error about trying to parse a NaN value, leaving the input empty (which, if they then clicked "Add to cart" I'm fairly sure would cause an error with Snipcart as well).

lpboucher: the type attribute does that for us. Also, it would make more sense to disable typing characters altogether by disabling the input

lpboucher (approved):

PR #77, opened 2021-07-07
Price-modifier-for-quantity_CU-nneqj3

a-wilkes:


a-wilkes:
Task linked: [CU-nneqj3 Price modifier for quantity](https://app.clickup.com/t/nneqj3)

lpboucher:
Task linked: [CU-nneqj3 Price modifier for quantity](https://app.clickup.com/t/nneqj3)

lpboucher:
Task linked: [CU-nhbe2j Add quantity input to product](https://app.clickup.com/t/nhbe2j)

a-wilkes:
Ok, so this ended up being more of a change that I think you were intending, so an explanation of my thought process is the following:
- Product price was a string, and the volume discount was parsing it then formatting it back to a string
    - This struck me as a violation of "single responsibility" as now the `VolumeDiscountPipe` was also responsible for ensuring prices were valid numbers and that they were formatted to two decimal places
- I added a method to get price as a number from product so now product was responsible for its own price
- I converted `VolumeDiscountPipe` to just operate on numbers, since that's all it should be doing
- When adding the total price pipe I realised that I'd also forgotten to put the euro symbol on the price for the featured product card, so this indicated that we might need a pipe to format the prices to two decimal places and also with the correct currency symbol

I hope that makes sense. I made the changes here as they all seemed to be connected—I couldn't really add the total price calculation without duplicating the parsing in the `VolumeDiscountPipe`, for example, so I needed the pipe to operate on numbers, and it just expanded out from there.

Only other note is that we'll need to remember to add the `totalPrice` method to the abstract product class when that gets merged in to develop.

lpboucher (approved):

lpboucher:
That makes a lot of sense, I like the initiative.

I still have to decide whether we store price as a string or as a number in the db. So the changes are good and we just adjust accordingly.

Yes, will fix the totalPrice when merging

PR #74, opened 2021-07-06
Add-quantity-input-to-product_CU-nhbe2j

a-wilkes:

PR #78, opened 2021-07-07
Product-detail-default-options_CU-nner2e

a-wilkes:


a-wilkes:
Task linked: [CU-nner2e Product-detail default options](https://app.clickup.com/t/nner2e)

lpboucher:
Task linked: [CU-nner2e Product-detail default options](https://app.clickup.com/t/nner2e)

lpboucher:
Task linked: [CU-nhbe2j Add quantity input to product](https://app.clickup.com/t/nhbe2j)

a-wilkes:
I branched this from [Add-quantity-input-to-product](https://github.com/lpboucher/dak-coffee-wholesale/pull/74) since it already had the forms and controls set up which this PR needed to change/subscribe to changes to.

lpboucher:
Got it. That's fine.

We should probably also merge develop into this branch and then change the destination branch to develop given that I merged a lot of tickets this weekend

lpboucher (changes requested):
I think this is quite extensive and can likely be simplified.

I will update the ticket with more information. Maybe as an indication, we will use the directive (from the tutorial) that we had looked at previously

src/app/shared/directives/bind-query-params.directive.ts
```diff
@@ -0,0 +1,62 @@
...
+
+    private formEqualsParams(formValues: any, paramValues: any): boolean {
+        const formEntries = Object.keys(formValues);
+        const paramEntries = Object.keys(paramValues);
+
+        return formEntries.length == paramEntries.length
+            && formEntries.reduce((res: boolean, key: string) => {
+                return formValues[key] == paramValues[key] && res;
+            }, true);
+    }
```
lpboucher: I find this method a bit clunky. reduce is not correct operator here. Here are two solutions:

1) .every: use .every on the array returned from Object.keys and compare the values against the other object's values. If it is true one way it is also true the other way. .every returns a boolean so it fits nicely here
2) When trying to check deep equality of objects, you can use toJSON and compare the resulting string. The caveat is that property order is important so if we thing that order of properties is not guaranteed then we would have to sort the keys

Also, it is likely that this type of equality check will be required at some later point. I have moved the logic to a utils/helper file

lpboucher (approved):

PR #80, opened 2021-07-12
Weight-pipe_CU-nxa7hf

a-wilkes:


a-wilkes:
Task linked: [CU-nxa7hf Weight pipe](https://app.clickup.com/t/nxa7hf)

lpboucher:
Task linked: [CU-nxa7hf Weight pipe](https://app.clickup.com/t/nxa7hf)

src/app/shared/pipes/weight.pipe.ts
```diff
@@ -0,0 +1,20 @@
+import { Pipe, PipeTransform } from "@angular/core";
```
lpboucher, line 1: this is perfect. Have some minor comments

src/app/shared/pipes/weight.pipe.ts
```diff
@@ -0,0 +1,20 @@
...
+
+import { Weight } from "@shared/models/types/weight.type";
+
+@Pipe({
+    name: "weightToGrams"
+})
+export class WeightPipe implements PipeTransform {
+    transform(weight: Weight): number {
+        const isGrams = !weight.includes("kg");
```
lpboucher: the base case is grams so I would prefer to use isKilos = weight.includes("kg")

also feels likeit reads a bit better

src/app/shared/pipes/weight.pipe.ts
```diff
@@ -0,0 +1,20 @@
...
+        const rawWeightValue = weight.split(/kg|g/)[0];
+
+        let weightValue = Number.parseInt(rawWeightValue);
```
lpboucher: here, we should use const. I get that you used let because you were reassigning the variable.

instead, we can do it in the return statement using the ternary operator (isKilos ? weightValue : weightValue / 1000)

The change is not necessarily to shorten the number of lines, but to avoid let as much as possible

lpboucher (changes requested):

lpboucher (approved):

PR #81, opened 2021-07-12
Product-weight_CU-nxa7gq

a-wilkes:


a-wilkes:
Task linked: [CU-nxa7gq Product weight](https://app.clickup.com/t/nxa7gq)

src/app/core/cart/cart.service.ts
```diff
@@ -61,12 +66,16 @@ export class CartService {
 
     private evaluateCartWeight(items: any): number {
         return items
-            .filter((item: any) => item?.dimensions?.weight != null && item?.quantity != null)
-            .reduce((sum: number, item: any) => sum + (item.dimensions.weight * item.quantity), 0);
+            .map((item: any) => item.quantity * this.getSnipcartItemWeight(item))
+            .reduce((sum: number, weight: number) => sum + weight, 0);
```
lpboucher: both map and reduce loop through the whole array. Chaining both results in O(2n) complexity.

We can do all the logic within a single reduce block to limit the complexity to O(n)

src/app/core/cart/cart.service.ts
```diff
@@ -61,12 +66,16 @@ export class CartService {
...
+    }
+
+    private getSnipcartItemWeight(item: any): number {
+        let weightField = item.customFields.find((field: any) => field.name == "Weight");
```
lpboucher: weightField is never reassigned so we should use const here.

Also, except when checking for null/undefined, we should always use strict equality: field.name === "Weight".

src/app/core/cart/cart.service.ts
```diff
@@ -61,12 +66,16 @@ export class CartService {
...
+        return weightField?.value != null ? this.weightPipe.transform(weightField.value) : 0;
```
lpboucher: are there instances when the value will be null when weightField is not null?

I am not sure, but I think just doing weightField ? this.weightPipe... : 0 is sufficient. I am not 100%, there may be some edge case that I am not thinking of

src/app/core/pricing/pricing-tier.service.ts
```diff
@@ -23,7 +23,7 @@ export class PricingTierService {
         this.isVolumeDiscountActive$.next(value ?? !this.isVolumeDiscountActive$.value);
     }
 
-    updateDiscount(itemCount: number): void {
-        this.toggleDiscount(itemCount > ITEM_COUNT_THRESHOLD);
+    updateDiscount(cartWeight: number): void {
```
lpboucher, line 26: This is good. Maybe one thing I did not specify, we only programmatically toggle the discount when going from 30% to 45%, but no the other way around.

It may seem counterintuituive, but the logic is that weight may go down temporarily when people decide what to order and end up adding, removing, adding, etc. So we want to cover the case where people order above the threshold when they thought they would order for 30%, but give them the benefit of the doubt that when the toggle gets activated, it should not get deactivated. (this will of course get further managed on our side when creating invoices)

lpboucher (changes requested):

lpboucher (approved):

PR #85, opened 2021-07-14
Product-options-price-modifiers_CU-nxa4xh

a-wilkes:


a-wilkes:
Task linked: [CU-nxa4xh Product options price modifiers](https://app.clickup.com/t/nxa4xh)

lpboucher:
Task linked: [CU-nxa4xh Product options price modifiers](https://app.clickup.com/t/nxa4xh)

a-wilkes:
I tried to make the changes here be as simple as possible since whilst making them I realised we have a much bigger issue which might require far more extensive changes.

Essentially, the product types are getting a bit mixed up. For example, all the product cards now show options for weight and roast, including the merchandise, which isn't right at all. In the current state we'll need to add an `ngIf` to the `product-card` and `product-detail` HTML to only optionally include this when the product is a coffee. However, a lot of the code related to weight and roast is found in the `base-product` which means all of this will need to be optionally wrapped as well.

This extends to this PR with the price modifiers for weight being essentially constants, but which require specific methods to be called. So, the base weight of 250g has a modifier of 0, which is no problem, but 1kg requires some code somewhere to call `kgPriceAsNumber` on a coffee object, but you can't add a method call to a constant definition and we currently only interact with the `Coffee` subclass as a generic `Product`, so we'd need to add some logic somewhere to check the type (which is a little messy since it's logic that merchandise won't use). Perhaps there's some solution where we move these definitions to be part of the `coffee` class, so the `coffee` class points at the constant definitions and can return everything connected together.
```ts
export class Coffee extends Product {
    //...
    get weightOptions() {
        return [
            {
                list: WEIGHT_OPTIONS[0],
                price: 0,
            },
            {
                list: WEIGHT_OPTIONS[1],
                price: this.kgPriceAsNumber(),
            },
        ]
    }
}
```
I think the problem with this, though, is that it becomes very easy to add a new weight class but forget to modify this method, which is a little ugly (and it still doesn't really solve the problem of coffee and merchandise logic getting mixed up).

I'm not really sure how to tackle this, and it's not strictly relevant to this PR (though this PR extends the problem a little, I think), so it's up to you if you want to create a new ticket (if you've some idea of how to solve it), or if you want me to just go through and add a bunch of conditionals to stop merchandise showing all the coffee options.

lpboucher:
Ahh I see your point. Adding conditionals all over the place is definitely not an option.

I will request changes and update the ticket with some indication of how to proceed

lpboucher (changes requested):
Will adjust ticket

a-wilkes:
`WeightPipe` still uses the `Weight` type (and `WeightPipe` is used in `CartService` in develop) and `SampleConfirmationModal` still uses the `Roast` type, so I didn't delete them just yet.

`SampleConfirmationModal`'s use of `Roast` could be replaced with just a `string` array, but this would require us to remember to update it if things ever changed in the `Product` subclasses.

Likewise, `WeightPipe`'s use of `Weight` could be replaced with just a `string`, though we'd need to be more thorough with error checking as there would no longer be a means to type-check the input.

Let me know how you want to proceed.

src/app/modules/products/components/product-card/product-card.component.html
```diff
@@ -6,13 +6,17 @@
         <span class="product-card-body-highlight">New</span>
         <h2 class="product-card-body-title">{{ product.name }}</h2>
         <form class="product-card-body-selection-form" [formGroup]="optionsForm">
-            <app-dropdown class="product-card-body-selection-item" formControlName="weight" [options]="weightOptions" [defaultSelection]="defaultWeight"></app-dropdown>
-            <app-dropdown class="product-card-body-selection-item" formControlName="roast" [options]="roastOptions" [defaultSelection]="defaultRoast"></app-dropdown>
+            <app-dropdown
+                class="product-card-body-selection-item"
+                *ngFor="let attribute of productAttributes"
```
lpboucher: unnecessary getter, you should take advantage of the rendering loop instead of adding additonal loops.

Here we can use let attribute of product.attributes

src/app/modules/products/components/product-card/product-card.component.html
```diff
@@ -6,13 +6,17 @@
...
+                formControlName="{{ attribute }}"
```
lpboucher: following the comment above, the formControlName becomes attribute.name

src/app/modules/products/components/product-card/product-card.component.html
```diff
@@ -6,13 +6,17 @@
...
+                [options]="getAttributeOptions(attribute)">
```
lpboucher: this getter can also be simplified by passing attribute.options instead of the whole attribute object

src/app/modules/products/components/product-detail/product-detail.component.html
```diff
@@ -12,24 +12,17 @@ <h1 class="product-detail-content-title">{{ product.name }}</h1>
...
-                        <label class="product-detail-selection-label" *ngFor="let option of weightOptions">
-                            <input formControlName="weight" type="radio" value="{{ option }}" class="product-detail-selection-input" />
-                            {{ option }}
-                        </label>
-                    </div>
-                    <div class="product-detail-selection-group">
-                        Roast:
-                        <label class="product-detail-selection-label" *ngFor="let option of roastOptions">
-                            <input formControlName="roast" type="radio" value="{{ option }}" class="product-detail-selection-input" />
+                    <div class="product-detail-selection-group" *ngFor="let attribute of productAttributes">
```
lpboucher: same as above

src/app/shared/abstracts/base-product/base-product.component.ts
```diff
@@ -1,62 +1,35 @@
...
         return this.optionsForm.get("quantity")!.value;
     }
 
-    get roastOptions(): string[] {
-        return this._roastOptions.map(r => r as string);
-    }
-
-    get weightOptions(): string[] {
-        return this._weightOptions.map(w => w as string);
+    get productAttributes() {
```
lpboucher: not required

We are already looping through the product in the hmtl file

src/app/shared/abstracts/base-product/base-product.component.ts
```diff
@@ -1,62 +1,35 @@
...
-                list: this.roastOptions,
-                selection: this.roast,
-            },
-        ];
+    get cartModifiers(): CartModifier[] {
+        return this.product
+            .attributes
+            .map(
+                attr => {
+                    return {
```
lpboucher: you can use an implicit return also with objects. Simply use a leading (

.map(attr => ({})

src/app/shared/abstracts/base-product/base-product.component.ts
```diff
@@ -66,4 +39,36 @@ export abstract class BaseProductComponent {
...
+                        [attr.name]: [attr.options[0].optionName, Validators.required],
+                        ...obj
+                    }
+                }, {}
+            );
+    }
+
+    getAttributeOptions(attribute: string): string[] {
+        const attr = this.product
```
lpboucher: here, we should loop through product attributes in the html file, that way we would only access one given attribute at a time, which would prevent having to use .find.

Also, it would greatly simplify because we can simply pass an options array (instead of the whole attribute object) and map through names

src/app/shared/models/classes/product-attribute.class.ts
```diff
@@ -0,0 +1,11 @@
+import { ProductAttributeOption } from "@shared/models/types/product-attribute-option.interface"
+
+export class ProductAttribute {
+    name: string;
+    options: ProductAttributeOption[];
+
+    constructor(name: string, options: ProductAttributeOption[]) {
```
lpboucher: please use the same structure as other classes (using ProductAttributeShape) so that we can init the class using an object instead of a list of parameters

src/app/shared/directives/snipcart-add.directive.ts
```diff
@@ -37,17 +41,16 @@ export class SnipcartAddDirective implements OnInit, OnChanges {
...
-        });
-    }
-
-    private rawListToSnipcartList(raw: string[]): string {
-        return raw.reduce((prev, curr) => prev + "|" + curr);
+        this.modifiers
+            .forEach(
+                (modifier, index) => {
+                    const basename = `data-item-custom${ index + 1}`;
+                    const optionsList = this.attributeToSnipcartListPipe.transform(modifier.attribute);
```
lpboucher: There is some inconsistency in the variables, methods etc.

You can the variable optionsList, but pass an attribute to the pipe

src/app/shared/pipes/attribute-to-snipcart-list.pipe.ts
```diff
@@ -0,0 +1,14 @@
+import { Pipe, PipeTransform } from "@angular/core";
+
+import { ProductAttribute } from "@shared/models/classes/product-attribute.class";
+
+@Pipe({
+    name: "attributeToSnipcartList"
+})
+export class AttributeToSnipcartListPipe implements PipeTransform {
+    transform(attribute: ProductAttribute): string {
```
lpboucher: the pipe only uses the nested options array of the passed attribute. Instead, you should pass the options array directly.

Also, reduce tends to reduce readability so should be used as a last resort.

This pipe can be cleaned by using map(opt) => `${opt.optionName}[${ o.priceModifier }]`

and then returning array.join("|")

src/app/modules/products/components/product-card/product-card.component.html
```diff
@@ -6,13 +6,17 @@
...
+            </app-dropdown>
             <app-quantity-input formControlName="quantity"></app-quantity-input>
         </form>
         <div class="product-card-body-price-container">
             <div class="product-card-body-price">{{ totalPrice | volumeDiscount | formatPrice }}</div>
```
lpboucher: this price does not update based on changes in the dropdown

lpboucher (changes requested):
I think we can simplify this a bit in terms of readability, performance

PR #86, opened 2021-07-14
Reactivate-filters_CU-mx91hm

a-wilkes:


a-wilkes:
Task linked: [CU-mx91hm Reactivate filters](https://app.clickup.com/t/mx91hm)

src/app/shared/pipes/filter.pipe.ts
```diff
@@ -1,26 +1,26 @@
...
+    includedByFilter(filter: ActiveFilters, item: Object): boolean {
         return Object.entries(item)
             .some(([k, v]) => this.matchingValue(filter, k, v));
     }
 
-    matchingValue(filter: FilterType, key: string, value: string): boolean {
+    matchingValue(filter: ActiveFilters, key: string, value: string): boolean {
         return Object.keys(filter).includes(key)
-            && filter[key].subOptions.includes(value);
+            && filter[key].includes(String(value));
```
a-wilkes, line 24: For some reason, although `value` is of type `string`, when the prices were getting passed in (also of type `string` on the `Product` class), they were getting converted to `number` types, so the `includes` call was then failing as, for example, `"12.25" == 12.25` is `false`. Explicitly converting to `string` fixed this, but it doesn't feel like this should be necessary.

src/app/utils/factories/filters.ts
```diff
@@ -0,0 +1,71 @@
+import { Coffee } from "@shared/models/classes/coffee.class";
+import { FilterType } from "@shared/models/types/filter-type.type";
+import { Merchandise } from "@shared/models/classes/merchandise.class";
+import { Product } from "@shared/models/classes/product.class";
+import { ProductType } from "@app/shared/models/types/product-type.type";
+
+export function getFilterableProperties(products: Product[], productType: ProductType): FilterType {
```
a-wilkes: Without a factory function, code was getting spread out all over the place in kind of an ugly way. Getting to this factory function is what took me the most time. Hopefully, it should now be the case that all of the code related to products and their filterable properties is in the one place, instead of across several files.

In addition to this, I wasn't sure which properties you wanted to filter on, so I just added all of them. The good thing about this solution, is that you can add and remove what you want from here, and you shouldn't need to worry about changing anything anywhere else. It should (hopefully) just work.

src/app/modules/products/pages/product/product.page.html
```diff
@@ -9,10 +9,13 @@
             </ng-template>
         </ng-container>
     </app-grid-list>
+    <ng-container *ngIf="products$ | async as products">
+        <app-filter [filterOptions]="getFilterableProperties(products)" (selectedOptions)="onSelectionChange($event)"></app-filter>
```
a-wilkes: I wasn't sure how you wanted the filter styled, so I've just left it as it comes out the box at the moment. Let me know if you want it styled in a specific way.

lpboucher: That's fine, we can adjust later. They should definitely be above all other product lists (featured and regular)

src/app/modules/products/pages/product/product.page.html
```diff
@@ -9,10 +9,13 @@
             </ng-template>
         </ng-container>
     </app-grid-list>
+    <ng-container *ngIf="products$ | async as products">
```
lpboucher: you should not have to use the async pipe for the same observable twice, otherwise it creates two separate subscriptions which hogs unnecessary additional memory.

Instead, your ng-container should wrap blocks that will use the subscribed observable and then as children you can use let products for the product list, but also let products for the filters

src/app/shared/components/filter/filter.component.ts
```diff
@@ -51,28 +75,30 @@ export class FilterComponent implements OnInit {
...
 
-        this.router.navigate([], { queryParams: filter });
+        this.selectedOptions.emit(filter);
+        this.router.navigate(
+            [],
+            {
+                relativeTo: this.activatedRoute,
+                queryParams: filter,
+            }
+        );
```
lpboucher, line 113: is there a specific reason why we are emitting and updating query params?

If it is not overly complicated we should use the query params to reflect the active filters so the components are fully decoupled

src/app/shared/pipes/filter.pipe.ts
lpboucher: that's fine for now, we can always convert price to a number in the class, might be easier

src/app/utils/factories/filters.ts
lpboucher: This is a good way to think about it. 

However, instead of using a generic factory function that switches on productType, we can use the classes that are extended based on productType as there is already the discinction.

I have updated product-option modifiers ticket and will update this ticket as well

lpboucher (changes requested):
I think there are a few things we worked on that can simplify this ticket a lot.

I will update the ticket

src/app/modules/products/pages/product/product.page.ts
```diff
@@ -12,27 +19,90 @@ import { Product } from "@shared/models/classes/product.class";
...
-        this.subscriptions.add(this.route.params
-            .subscribe(
-                ({productType}) => {
-                    this.products$ = this.productService.getProductsByType(productType);
-                }
+        this.subscriptions.add(
+            this.route.params.subscribe(
+                ({productType}) =>
+                    this.subscriptions.add(
+                        this.productService.getProductsByType(productType).subscribe(
```
lpboucher: we should not use nested subscriptions. RxJS has operators suchs as switchMap, forkJoin etc. to combine multiple observables into a single one

src/app/modules/products/pages/product/product.page.ts
```diff
@@ -12,27 +19,90 @@ import { Product } from "@shared/models/classes/product.class";
...
     ngOnInit(): void {
         this.featuredProducts$ = this.productService.getFeaturedProducts();
 
-        this.subscriptions.add(this.route.params
-            .subscribe(
-                ({productType}) => {
-                    this.products$ = this.productService.getProductsByType(productType);
-                }
+        this.subscriptions.add(
+            this.route.params.subscribe(
```
lpboucher, line 42: Can we not use snapshot here? I know we are updating params in the component but we are updating query params, not route params

src/app/modules/products/pages/product/product.page.ts
```diff
@@ -12,27 +19,90 @@ import { Product } from "@shared/models/classes/product.class";
...
     }
+
+    private updateFilterForm(products: Product[], productType: ProductType): void {
+        this.filterableProperties
+            ?.forEach(p => this.filterForm.removeControl(p.key));
+
+        this.filterableProperties = (this.getFilterableProperties(products, productType));
+
+        this.filterableProperties
+            .forEach(p => this.filterForm.addControl(p.key, new FormControl()));
```
lpboucher: multiple forEach in the same block

src/app/modules/products/pages/product/product.page.ts
```diff
@@ -12,27 +19,90 @@ import { Product } from "@shared/models/classes/product.class";
...
+                    [fa.attribute]: {
+                        displayName: fa.displayName,
+                        key: fa.attribute,
+                        options: getUniqueValuesOfKey(products, fa.attribute),
+                    }
+                }
+            })
+            .reduce((union, obj) => {
+                return { ...union, ...obj }
+            });
```
lpboucher: using map, reduce, map, reduce means we are looping over the data 4 times.

src/app/modules/products/pages/product/product.page.ts
```diff
@@ -12,27 +19,90 @@ import { Product } from "@shared/models/classes/product.class";
...
+            ?.forEach(p => this.filterForm.removeControl(p.key));
+
+        this.filterableProperties = (this.getFilterableProperties(products, productType));
+
+        this.filterableProperties
+            .forEach(p => this.filterForm.addControl(p.key, new FormControl()));
+    }
+
+    private getFilterableProperties(products: Product[], productType: ProductType): FilterType[] {
+        const filters = products
```
lpboucher: transforming products and productType into filters is a good use case for a custom pipe

src/app/modules/products/pages/product/product.page.ts
```diff
@@ -12,27 +19,90 @@ import { Product } from "@shared/models/classes/product.class";
...
             )
         );
     }
 
     ngOnDestroy(): void {
         this.subscriptions.unsubscribe();
     }
+
+    private updateFilterForm(products: Product[], productType: ProductType): void {
+        this.filterableProperties
```
lpboucher: rather than looping through all filterableProperties. We should extract the new filterable properties, check if the key exists and remove/add the existing keys

src/app/modules/products/pages/product/product.page.ts
```diff
@@ -12,27 +19,90 @@ import { Product } from "@shared/models/classes/product.class";
...
+    private updateActiveFilters(changes: any): void {
+        this.activeFilters = Object.keys(changes)
+            .filter(key => changes[key] != null)
+            .map(key => {
+                const active = [...changes[key]];
+                return active.length == 0 ? {} : { [key]: active };
+            })
+            .reduce((union, obj) => {
+                return { ...union, ...obj }
+            }, {});
```
lpboucher, line 80: you can remove the filter part and add a check for !=null in the map portion of this block.

This is also a good place to use a getter/setter combination

src/app/shared/models/classes/coffee.class.ts
```diff
@@ -12,6 +12,13 @@ export class Coffee extends Product {
         super(coffeeShape);
 
         if (coffeeShape != null) {
+            this.filterableAttributes = this.filterableAttributes.concat([
```
lpboucher: why use concat here? Can you not do just this.filterableAttributes = [...]?

Okay, I see that you are also setting this property in the parent class. You should use the spread operator instead of concat

src/app/shared/models/types/filterable-attribute.type.ts
```diff
@@ -0,0 +1,4 @@
+export type FilterableAttribute = {
+    attribute: string,
```
lpboucher: we can derive this type from FilterType using Omit<FilterType, options>

src/app/shared/pipes/filter.pipe.ts
```diff
@@ -1,26 +1,26 @@
...
 @Pipe({
     name: "filter"
 })
 export class FilterPipe implements PipeTransform {
 
-    transform(items: Object[], filter: FilterType): Object[] {
+    transform<T>(items?: T[], filter?: ActiveFilters): T[] {
         if (items == null) { return []; }
-        if (filter == null) { return items; }
+        if (filter == null || Object.keys(filter).length == 0) { return items; }
```
lpboucher: use full equality check on primitives (.length === 0)

src/app/utils/helper.ts
```diff
@@ -9,3 +9,12 @@ export const simpleDeepEqual = (firstObject: { [key: string]: any }, secondObjec
...
     return haveSameNumberOfKeys && haveSameValues;
 }
+
+export function getUniqueValuesOfKey(objects: { [key: string]: any }[], key: string): any[] {
+    const values = objects
+        .filter(o => key in o)
+        .reduce((prev: any[], curr) => prev.concat(curr[key]), [])
+        .filter(value => value != null);
+    const uniqueValues = new Set(values);
+    return [...uniqueValues];
```
lpboucher: you should be able to do this in a single loop:

`return [...new Set(objects.map(item => item.key))];` and adjusting for missing keys, null values

lpboucher (changes requested):
it all works quite well, nice job!

Let's work on refactoring. You should always stay away from nesting .map, .reduce, .filter because they loop through full arrays so adding additional loops will affect performance quite a bit.

When adding methods within a component, sometimes it is easier to add a simple method that works on a single item and then call that method for all items in an array. It is generally more robust than passing arrays

src/app/modules/products/pages/product/product.page.ts
a-wilkes: I gave this a shot, but TS requires that the return value from a getter is exactly the same as the input value for a setter, so I couldn't see how to break this in to two parts. You'd either need to require all the callers of setters to process the input before they passed it in (which would be, essentially, this method here), or do everything in the setter, but lose type checking by having to return `any` (the type of `changes` from the form), and have to do a lot more error checking.

If I've misunderstood, let me know,

PR #87, opened 2021-07-19
Email-async-validator_CU-p7gyue

a-wilkes:


a-wilkes:
Task linked: [CU-p7gyue Email async validator](https://app.clickup.com/t/p7gyue)

src/app/modules/authentication/components/request-access-form/request-access-form.component.html
```diff
@@ -3,14 +3,15 @@
         <div class="request-access-card">
             <h1 class="request-access-title">Request Access</h1>
 
-            <form [formGroup]="requestAccessForm" (ngSubmit)="onSubmitRegistration()" class="request-access-form">
+            <form [formGroup]="requestAccessForm" (ngSubmit)="submit.focus(); onSubmitRegistration()" class="request-access-form">
```
a-wilkes, line 6: Without moving focus somewhere else, if the user has a different error displayed, then the form won't tell them what's actually wrong on submission when hitting `enter`, it will just continue to show the same error as before (e.g. it might continue to say "Please enter an email address" when they've already added one).

lpboucher (approved):

PR #89, opened 2021-07-26
Base-Product-refactor_CU-n7dy9j

a-wilkes:


a-wilkes:
Task linked: [CU-n7dy9j Base Product refactor](https://app.clickup.com/t/n7dy9j)

a-wilkes:
I think this should have all the relevant changes from the previous PR. A lot of the original changes were about the weight and roast options—which we have since changed for a more generic approach—so are now redundant. Hopefully, I've caught everything else from the previous PR, though.

PR #91, opened 2021-07-27
Wallet-discount_CU-pnfkua

a-wilkes:


a-wilkes:
Task linked: [CU-pnfkua Wallet discount](https://app.clickup.com/t/pnfkua)

src/app/core/authentication/authentication.service.ts
```diff
@@ -35,6 +38,7 @@ export class AuthService {
             this.router.navigate(["/"], { queryParams: { checkPricing: true } });
 
             this.pricingTierService.calculateWalletAmount();
+            this.cartService.applyDiscount();
```
a-wilkes, line 41: I called the cart service here instead of in the pricing tier service since the CS already has the PTS injected, so injecting the CS into the PTS as well results in a circular dependency.

a-wilkes:
I'm getting a 400 error code in response to setting the discount, specifically these two errors:
```
zone-evergreen.js:2863 POST https://app.snipcart.com/api/cart/537de9b7-9803-4336-b451-a2519e8861bc/discounts 400
```
```
ERROR Error: Uncaught (in promise): Object: {"kind":"validation","form":[],"fields":{"code":[{"validation":"promocode_isnt_applicable","message":"promocode_isnt_applicable"}]}}
    at resolvePromise (zone-evergreen.js:1213)
    at zone-evergreen.js:1120
    at t.executeOutcome (snipcart.js:1)
    ...
```

Based on the details in the second message, I'm assuming this is an issue with the discount code not being recognised, as opposed to my misuse of the API. Though, of course, it could be the latter.

PR #94, opened 2021-07-27
Footer-refactor_CU-ptc49w

a-wilkes:


a-wilkes:
Task linked: [CU-ptc49w Footer refactor](https://app.clickup.com/t/ptc49w)

a-wilkes:
I didn't extend from the abstract navigation class since that would only necessary if you also want the widgets included in the footer, and it wasn't clear to me that this was the case. As always, I can adjust if needed.

PR #96, opened 2021-07-27
Dak-Logo_CU-ptcdq7

a-wilkes:


a-wilkes:
Task linked: [CU-ptcdq7 Dak Logo](https://app.clickup.com/t/ptcdq7)

a-wilkes:
I saved the logo from your main site as an svg image and used it in an `img` element since we'll need it in that format when we merge this with the angular-svg-icon changes.

Also, things are getting a bit packed in the header once again, so we'll need to think about how to arrange everything again.

PR #98, opened 2021-07-28
Product-details_CU-ptc1v1

a-wilkes:


a-wilkes:
Task linked: [CU-ptc1v1 Product details](https://app.clickup.com/t/ptc1v1)

src/app/shared/components/slide-open-menu/slide-open-menu.component.scss
```diff
@@ -0,0 +1,42 @@
...
+    @apply overflow-hidden;
+
+    max-height: 0;
+    transition: max-height 500ms ease-out;
+}
+
+.show-content {
+    max-height: 100px;
+    transition: max-height 200ms ease-in;
+}
```
a-wilkes, line 42: I realise now just how much of a pain it is to get a sliding menu like this working. I couldn't get it even remotely working with tailwind so I just used pure CSS. I also set the target height to 100px, which is obviously problematic if the menu ends up taller. I really wasn't sure how else to solve the problem.

